PR #16, opened 2019-05-31
Add article on Gruntwork Philosophy

yorinasub17:
This publishes the Gruntwork Philosophy document that we have in our newer repos on the docs site so that it is available.

### DIFF
--- a/content/introduction/philosophy/index.md
+++ b/content/introduction/philosophy/index.md
@@ -1,0 +1,69 @@
+---
+title: "Gruntwork Philosophy"
+date: 2019-05-31
+---
+
+At Gruntwork, we strive to accelerate the deployment of production grade infrastructure by prodiving a library of
+stable, reusable, and battle tested infrastructure as code organized into a series of [packages](#what-is-a-package) with
+[modules](#what-is-a-module). Each package represents a particular set of infrastructure that is componentized into
+smaller pieces represented by the modules within the package. By doing so, we have built a composable library that can
+be combined into building out everything from simple single service deployments to complicated microservice setups so
+that your infrastructure can grow with your business needs. Every package we provide is built with the [production grade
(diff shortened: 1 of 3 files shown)


content/introduction/philosophy/index.md
```diff
@@ -0,0 +1,69 @@
+---
+title: "Gruntwork Philosophy"
+date: 2019-05-31
```
robmorgan: Should we always bump the date forwards to a future date?

content/introduction/philosophy/index.md
```diff
@@ -0,0 +1,69 @@
...
+---
+
+At Gruntwork, we strive to accelerate the deployment of production grade infrastructure by prodiving a library of
```
robmorgan:
```suggestion
At Gruntwork, we strive to accelerate the deployment of production grade infrastructure by providing a library of
```

content/introduction/philosophy/index.md
yorinasub17: I updated this because it is used in the `Last updated` marker. So we should update the date everytime we modify it so that that marker is accurate.

content/introduction/philosophy/index.md
yorinasub17: Good catch!

content/introduction/philosophy/index.md
robmorgan: Technically yes, but that's a different point. I was trying to say if you know you're going to publish this doc say later this week, set it to a future date such as `2019-06-05` so it looks 'fresh' when it goes live. WDYT?

yorinasub17: Ahhh I gotcha. Sorry I misunderstood. Yes that makes sense. We just have to be careful to time it right so that we don't have negative last modified timestamps.

yorinasub17:
Going to merge this and deploy now. I updated the timestamps to today so that they will look fresh when I deploy.

PR #28, opened 2019-07-01
Add terraform-google-vault

yorinasub17:
And syncs up all the indents.


yorinasub17:
Merging and deploying since content is straightforward.

PR #31, opened 2019-07-02
More tf12 compatibility

yorinasub17:
This adds all the repos that have been updated for tf12 compatibility that were previously not included before.


yorinasub17:
Merging and releasing given content. Will address any issues in a future PR if they are found.

PR #32, opened 2019-07-03
Added package-beanstalk

eak12913:


eak12913:
Ah - yes. It was my first time editing the file via GitHub directly. I've fixed it now.
Thanks for the 2nd set of eyes!

PR #33, opened 2019-07-05
More tf12 compatibility: `package-kafka` and `package-elk`

yorinasub17:
This adds `package-kafka` and `package-elk` to the list of tf12 compatible versions.


yorinasub17:
Merging and deploying given content.

PR #34, opened 2019-07-08
Add package-static-assets and separate GCP

yorinasub17:
This introduces the following two changes:

- Adds `package-static-assets` to the version compatibility chart.
- Separates out GCP modules from AWS modules in the version compatibility chart.


yorinasub17:
Thanks for the review! Merging and releasing.

PR #36, opened 2019-07-11
Add influx to tf12 compatibility list

yorinasub17:
As the title states, this adds the `terraform-aws-influx` module to the tf12 version compatibility chart.


yorinasub17:
Merging given content.

PR #38, opened 2019-07-23
Highlight the terraform-aws-eks module is private

yorinasub17:


josh-padnick:
Thanks for this prompt turnaround! This issue makes me realize the tension between a public docs site for non-public code repos, but perhaps it's at least worth more clearly setting expectations versus creating the impression that our docs are broken. To that end, does it make sense to:

1. Have a notice at the very top that says something to the effect of:

    ```
    Note that some links in this guide may not work unless you have an active Gruntwork subscription.
    ```

1. Put a small icon such as a lock icon next to any links that are private to Gruntwork subscribers?

1. Use a modal on any Gruntwork subscriber links that alerts users to the fact that they are about to click a link that will only work for subscribers (similar to our website).

I'm actually not a fan of any of these options but thinking out loud for now...

brikis98:
Note that I'll be working with Sam on these very questions for the Launchpad stuff we've been discussing. My plan was to have clear pointers for private links with a CTA explaining what those links are all about. Exact design TBD.

josh-padnick:
Circling back a few days later here, this update seems like an improvement over what we have now, so my vote is that we merge this as is. Later on, when we release more deployment guides, we can incorporate better ways of handling proprietary links.

So feel free to merge and deploy!

yorinasub17:
Ok sounds good. Will merge for now.

PR #40, opened 2019-08-07
Fix broken links on upgrade guide

yorinasub17:
Gatsby doesn't auto tag the headers, so we need to create anchors 😞


josh-padnick:
Does https://www.gatsbyjs.org/packages/gatsby-transformer-remark add the anchors you want?

yorinasub17:
We're actually already using that, and it doesn't add the anchors... 😞

PR #42, opened 2019-08-14
Add section listing known issues with upgrading the reference architecture

yorinasub17:
This adds a new section "Known Issues" to the upgrade guide, which highlights the issues with upgrading the tf11 reference architecture to tf12.

The specific issue listed here is what one of our customers were running into with the Reference Architecture.


content/guides/upgrading-to-tf12-tg19/index.md
```diff
@@ -349,3 +349,117 @@ here:
...
+
+    This value does not have any attributes.
+
+    ...
+    ```
+
+  This is due to a bug in terraform where it can not properly handle the migration of state representation from a single
+  resource to a resource using `for_each`. You can read more about the issue on [the bug ticket
+  here](https://github.com/hashicorp/terraform/issues/22375). To resolve this, you need to do state surgery to update
+  the representation. This will be done by first removing the original resource from the state, and then importing it.
```
brikis98: face, meet palm.

yorinasub17: agreed... 😞

yorinasub17:
Merging and deploying. Thanks for the review!

josh-padnick:
This is exactly the type of Gruntwork that makes customers lives easier. Great to see this captured, Yori.

PR #44, opened 2019-08-29
Add GCP Reference Architecture Page

robmorgan:
This PR adds a placeholder page about the upcoming GCP Reference Architecture.

### DIFF
--- a/content/reference/gcp-reference-architecture/index.md
+++ b/content/reference/gcp-reference-architecture/index.md
@@ -1,0 +1,25 @@
+---
+title: "GCP Reference Architecture"
+date: 2019-08-29
+---
+
+We are building out an end-to-end, production-grade, secure, and developer-friendly Reference Architecture for GCP. Just as with our [AWS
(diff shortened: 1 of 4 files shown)


robmorgan:
Merging given docs

josh-padnick:
Rob, great to have something official we can point clients to on the GCP RefArch. My main feedback was to make this offering more concrete by describing it in the present tense ("we include") vs. future tense ("we will include").

Separately, the public website is probably the better place for this, but since we're in the process of re-working that, this is a good start.

robmorgan:
> Rob, great to have something official we can point clients to on the GCP RefArch. My main feedback was to make this offering more concrete by describing it in the present tense ("we include") vs. future tense ("we will include").

Thanks for the feedback, fixed in: https://github.com/gruntwork-io/docs/commit/b177013b35fb6503ce00d54cad414ae5fd0e1088

> Separately, the public website is probably the better place for this, but since we're in the process of re-working that, this is a good start.

I agree it just made sense to put this somewhere and start linking to it in customer emails.

PR #57, opened 2021-11-24
[APT-1570] [APT-1572] [APT-1555] [APT-1581] Bootstrapping nav and layout

ebeneliason:
This PR includes work across several of my stories. I've broken it down into logical chunks by commit, so I recommend reviewing one commit at a time to make sense of each change. That said, I felt that seeing it all together in the preview was more useful than viewing piecewise, hence the single PR.


eak12913:
Should we consider having the "light" background be the same as in Aperture?
image

ie - not _white_ but just slightly off-white?

eak12913:
From Ore:
image
There appears to be some extra spacing after the last item in the sublist..do we want to keep the left border line when there's empty space like that? It may make it seem like something is supposed to be there but not rendering.

docusaurus.config.js
```diff
@@ -46,10 +51,11 @@ const config = {
     /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
     ({
       navbar: {
-        title: "Gruntwork Docs",
+        title: "",
```
eak12913, line 54: We're getting rid of the title?

eak12913: Oh perhaps it's logo now that's entirely an image?

src/css/custom.css
```diff
@@ -7,39 +7,41 @@
 /* You can override the default Infima variables here. */
 :root {
   /* default */
-  --ifm-color-primary: #25c2a0;
+/*  --ifm-color-primary: #25c2a0;
```
eak12913, line 10: keep this?

eak12913 (approved):
Approving for merge. We agreed to clean up any loose ends as we make further progress.

PR #65, opened 2021-12-01
[APT-1564] Service Catalog reference content 

eak12913:
This PR adds content which describes all of the services currently available in our Service Catalog


_docs-sources/docs/reference/services/intro.md
```diff
@@ -0,0 +1,24 @@
+# Gruntwork Service Catalog for AWS
```
ebeneliason, line 1:
```suggestion
# Overview
```

ebeneliason (approved):
There are some known quirks with rendering of the inputs/outputs, but this was stretch for this story and we have another planned to clean it up, so we'll go with it as is for now! Looks great otherwise.

PR #68, opened 2021-12-02
[APT-1557] Bump version of docs-sourcer

eak12913:
Bump to the next version to source content out of `_docs-sources/` rather than `_docs-sources/docs`


ebeneliason (approved):
LGTM

PR #72, opened 2021-12-02
[APT-1567] Import Refarch content

oredavids:
[APT-1567]

[APT-1567]: https://gruntwork.atlassian.net/browse/APT-1567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eak12913:
Suggestion: Ditch the "Intro" chapters a la:
image

Remember - you thought that it looks odd when you are looking at the main introduction and then you have the "next" button at the bottom that just says "Intro" and we couldn't figure out a good thing to call that "intro" because we didn't want to just repeat the folder name..etc.. - Well I think ditching that whole intro section is a better solution because we already have the sidebar. It's showing you the outline. Why do we need a separate page. There's no real info on that page. It's essentially a bulleted list of the sub chapters. It makes sense in GitHub (where there's no sidebar) but my assertion is that we don't need it.

I think getting rid of that section will make the "next" button have a much more logical name: "Setting up initial access"

eak12913 (approved):
Approving and merging. We can tweak this further later

PR #79, opened 2021-12-07
[WIP][DO NOT MERGE] Raw guides migrated to MD

eak12913:
This PR is not to be merged - it contains all of the raw Guides that are taken from gruntwork.io and have been converted from AsciiDoc into Markdown.

Unfortunately, the preview environments are currently not working because the converted MarkDown files have some HTML tables in them. These tables have `style` tags that are pure HTML. Docusaurus rendering engine appears to expect _JSX_ and React is complaining and failing the build. Locally, it seems to be possible to run `yarn start` to bring up the site and view the MarkDown files.

They are all located in: `/guides/migrated`. For example: `http://localhost:3000/docs/guides/migrated/2020-11-09-golang-style-guide`


ebeneliason:
Hmm, the style guides look decent at a glance, but I seem to be unable to load any of the longer documents. They take up to a minute to load (according to the progress bar), but then fail to ever actually render. Shouldn't they be effectively immediate to load despite their length?

eak12913:
I've seen this happen with NextJS before and it's super annoying. It looks like the progress bar takes forever when you navigate using a link. However, if you stop the progress of the page and then just hit `ENTER` in the address bar it will load instantly....

I have observed this on various pages made with NextJS (including the docs for next-auth)

PR #91, opened 2021-12-13
Toc single item fix

ebeneliason:


ebeneliason:
Note: we should merge this as 2 separate commits so we preserve the change we made to the swizzled file.

PR #93, opened 2021-12-13
Use cards in the home page layout

ebeneliason:
In addition to adding cards to the home page, this adds some addition features to our `Grid` component:
- Responsive
- Option to stack multiple grids together (with different column widths)
- Better support for equal height rows
- Shorthand `gap` property for equal row and column gaps

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -31,7 +31,7 @@ export default function Home(): JSX.Element {
   const { siteConfig } = useDocusaurusContext()
   return (
     <Layout
-      title={`Hello from ${siteConfig.title}`}
+      title={`Gruntwork Docs`}
       description="Documentation and guides for Gruntwork's tools and services"
     >
       <HomepageHeader />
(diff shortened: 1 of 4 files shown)


src/pages/index.tsx
```diff
@@ -1,40 +1,90 @@
...
-  );
+  )
 }
 
 export default function Home(): JSX.Element {
-  const {siteConfig} = useDocusaurusContext();
+  const { siteConfig } = useDocusaurusContext()
   return (
     <Layout
       title={`Hello from ${siteConfig.title}`}
```
eak12913, line 34:
```suggestion
      title={`Gruntwork Docs`}
```

src/pages/index.tsx
```diff
@@ -1,40 +1,90 @@
...
-      description="Description will go into a meta tag in <head />">
+      description="Description will go into a meta tag in <head />"
```
eak12913, line 35:
```suggestion
      description="Documentation and guides for Gruntwork's tools and services"
```

eak12913 (approved):
Ship it.

PR #96, opened 2021-12-14
Responsive fixes for home page

ebeneliason:
This tweaks the home page layout for better mobile support, including:

- Shorten tutorial button text to fit
- Hide the "Gruntwork Docs" header, which wraps and is redundant with logo
- Add some margin to the cards section
- Prevent the nav bar items from word-wrapping


oredavids (approved):
LGTM.

PR #99, opened 2021-12-15
[APT-1585] Courses page

ebeneliason:
This is a solid start towards the dedicated courses page. I've chosen to include the page underneath `/docs/guides` as a `.mdx` file, which enable us to utilize built-in layouts and markdown on the page in conjunction with the Courses component.


ebeneliason:
This is now ready for review. [Here's the direct link to the page](https://deploy-preview-99--pensive-meitner-faaeee.netlify.app/docs/courses/).

Note that there are some minor discrepancies between this and the mockup in Figma. Most notably, the page width and margins aren't the same. That's an artifact of utilizing the built-in Docusaurus layout, which we want to do since it allows us to utilize markdown (including the admonition) for the rest of the page. I'm okay with that tradeoff.

I briefly explored adjusting the style conditionally for pages without sidebars, but I wasn't able to account for presence or absence of the TOC in the selector, so it proved near-impossible to achieve.

eak12913 (commented):
It looks pretty darn good - but I wonder if we can get the content centered?

image

I see that you're hiding the the TOC - but I'm wondering if we need more of a centered layout for this page.

ebeneliason:
[Womp.](https://github.com/facebook/docusaurus/issues/737)

> Custom layout got removed because upon investigation it actually doesn't work 😭

> Let's leave custom layout on v2 implementation


Custom layouts don't appear to be available to us just yet. I would love to center it, but found no way to do so that wouldn't interfere with other pages, or result in an extra wide layout (when hiding the TOC).

ebeneliason:
We found a solution for centering by adding an MDX wrapper to the content.

eak12913 (approved):
LGTM!

PR #108, opened 2021-12-16
Tweak card styles for better handling of card content

ebeneliason:
The primary motivation here is to refine the layout of the cards
on the courses page, but some of the tweaks apply more generally.
This also corrects other styling on the courses page via
improvements to the <CenterLayout> component.


eak12913 (approved):
LGTM!

PR #119, opened 2022-01-03
[APT-1573] [APT-1574]: Subscriber notification for private Github links

oredavids:
This PR adds a modal popup that will display to all users when they click on a link that takes them to `github.com/gruntwork-io/*`. The modal will warn users about getting 404s if they are not subscribers. There's also the ability to indicate that the warning should never be shown again.


oredavids:
@eak12913 I noticed that when I clicked the checkbox(to prevent future notifications) and then hit the "Cancel Option"; I got the notification on the next link that I clicked. This is minor but shouldn't we always respect the notification preference?

src/components/Modal.tsx
```diff
@@ -0,0 +1,129 @@
...
+  showModal,
+  shouldCloseOnEsc = true,
+  shouldAcceptOnEnter = false,
+  shouldCloseOnOverlayClick = true,
+  handleCancelRequest,
+  handleAcceptRequest,
+}) => {
+  const onRequestClose = () => {
+    // If the user checked to never see this notice but subsequently cancels we will disregard their selection. We will
+    // only stop showing this notice if they check the box and then proceed to GitHub
```
oredavids, line 34: I see that this is intentional - Is there a particular reason for disregarding it?

eak12913: We can discuss this quickly next time we speak - but for the purpose of this PR we _do_ want to have that behavior. It's a slightly better user experience as it applies your wishes only after you've chosen to follow the link forward. Otherwise you can get into a s situation where you dind't mean to dismiss the thing and never see it again.

oredavids: 👍🏾

ebeneliason (approved):
In the future we should refactor this so that we have a generic modal that's styled for our purposes, but let's get this in as-is.

PR #124, opened 2022-01-04
[APT-1630] Fix Refarch guide lists numbering

oredavids:
[APT-1630]

[APT-1630]: https://gruntwork.atlassian.net/browse/APT-1630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_docs-sources/guides/reference-architecture/02-authenticate/01-intro.md
+++ b/_docs-sources/guides/reference-architecture/02-authenticate/01-intro.md
@@ -9,20 +9,20 @@ In the last section, you got a basic [overview of the architecture](../01-overvi
 a variety of infrastructure deployed across multiple AWS accounts. In this section, you'll learn about authenticating
 and connecting to all the resources in your AWS accounts:
 
-* [Set up initial access](setting-up-initial-access): If this is your first time using this infrastructure—that is,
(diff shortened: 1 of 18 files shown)


ebeneliason (commented):
As best I can tell, the changes in this PR fall into 3 categories:
1. Indenting code blocks which reside inside lists
2. Switching to use of `1.` for all items in numbered lists
3. Changing the link format of some links in list items

(1) seems straightforward. (2) shouldn't, as I understand it, actually have an impact on the output. Does it? Do we definitely want to adopt that convention universally? I don't understand the rationale for (3). Is that an incidental change, or does the format of the link change the way it gets processed in some way? If so, how would we make it obvious to authors what format to use?

oredavids:
@ebeneliason 
(2) The use of `1.` has no impact on output. I made the change because we had most of the numbered lists using that within the RefArch guide.

(3) I made a change to the link format so that it works in Github as well as Docusaurus. We made this change in other guides before the holidays but didn't do this for the Refarch because it was migrated earlier.

ebeneliason (approved):
Thanks for the clarifications on those changes. Looks good to me, although admittedly I didn't scour everything to be sure none were _missed_, so I'm trusting you/regex on comprehensiveness.

PR #125, opened 2022-01-04
[APT-1575] Add support for image captions via plugin

ebeneliason:
The plugin identifies image content in the page with alt text specified,
and injects a new emphasis node immediately following containing that text.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -5,6 +5,8 @@ const lightCodeTheme = require("prism-react-renderer/themes/github")
 const darkCodeTheme = require("prism-react-renderer/themes/dracula")
 const cfg = require("config")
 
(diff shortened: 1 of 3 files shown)


oredavids (approved):
Nice!

ebeneliason:
Thanks @oredavids! Would you mind re-stamping? I pushed a new commit with better comments to really explain how it all works for posterity.

oredavids (approved):

eak12913:
@ebeneliason  this is great - but is this a plugin that you wrote? Or is this something that we sourced from the official image caption plugin and possibly adapted to have proper imports? If we sourced it from the original - we should document that and explain why we had to write our own, when there's an image caption remark plugin already available.

PR #127, opened 2022-01-06
Revised home page layout

ebeneliason:
This tweaks copy and layout on the homepage to provide a better
visual hierarchy. It also introduces some changes to the Card
component and styles to support the new look in cleaner fashion.


josh-padnick:
I was just browsing this and this looks awesome! Some additional small feedback:

- **What is Gruntwork?** Shorten text to "<div>Learn how Gruntwork products can help you deploy a world class infrastructure.</div>"
- I love the RefArch vs "Deploy a Service" breakdown. The icons are great, too!
- **Setup your accounts.** Should this be "Setup Your AWS Multi-Account Structure"?
- **Create a CI/CD Pipeline.** Adjust to "Create a CI/CD Pipeline for Infra." Might be too long, but we want to clarify it's for infra.
- Change "deploy a k8s cluster" to "Deploy Kubernetes using EKS"?
- Finally, perhaps there's some kind of common icon we can use to signify a "use case"? Maybe a single cube, or a document icon or something else. 

These are all off-the-cuff ideas. You should make the final call on everything here.

src/components/Grid.tsx
```diff
@@ -34,7 +34,7 @@ export const Grid: React.FunctionComponent<GridProps> = ({
   rowGap = rowGap || gap
 
   return (
-    <section
+    <div
```
eak12913, line 37: This is more semantically correct right? `section` refers more to a "section" in a document rather than this..being a grid, which is more of a container. Am I understanding this properly?

src/pages/index.tsx
```diff
@@ -36,50 +35,73 @@ export default function Home(): JSX.Element {
...
+              >
+                Learn how Gruntwork products can help you deploy a world class
+                infrastructure.
+              </Card>
+              <Card
+                title="The Reference Architecture"
+                href="/docs/guides/reference-architecture/overview/overview"
+                icon="/img/icons/refarch.svg"
+              >
+                Bootstrap your infrastructure in about a day by letting
```
eak12913, line 52: Nit: This feels more like a marketing blurb and less like verbiage that will help customers find docs for interacting with their Ref Arch.

I could see something like: "Got a Ref Arch deployed? Learn all about it" <-- not this exact text...but the gist should be "if you got the ref arch, this is where you can learn about how to absorb it into your org"

eak12913 (approved):
I had a minor nit - but it's not a blocker. So I'm approving this and can re-stamp if we do end up making changes.

src/components/Grid.tsx
ebeneliason: I had the same thought initially, but in practice I noticed a couple of related things that changed my mind.
1. In some cases the Grid belongs logically grouped with some other content, e.g. a header, or supplemental text. In those cases, the semantic "section" is all of that content taken together, not _just_ the grid.
2. In some cases, the component using the grid already had a higher level notion of _section_, which resulted in nested sections. Nested sections aren't prohibited, but semantically it was again incorrect, since the section is meant to map to a section of the page that would be shown in a document outline.

src/pages/index.tsx
ebeneliason: Yeah, that's fair. Updated!

eak12913 (approved):
LGTM!

PR #139, opened 2022-01-18
[APT-1652] Fix alignment/zoom for avatars on courses page

ebeneliason:


docusaurus.config.js
```diff
@@ -228,7 +228,7 @@ const config = {
             contextualSearch: true,
           }
         : undefined,
-      zoomSelector: ".markdown img",
+      zoomSelector: ".markdown :not(em) > img:not(.no-zoom)",
```
eak12913, line 231: What's `em`?

ebeneliason: Am emphasis (`<em>`) tag. This selector is a little tricky, as it is designed to handle two separate cases in which we _don't_ want zoom.
1. The first is the `:not(em)` bit, which is a standard approach to preventing image zoom documented in some of the zoom plugins. It enables one to wrap the Markdown image syntax in underscores (`_`) to trigger an emphasis (`<em>`) wrapper, which thus makes this selector fail.
2. The second adds an explicit `.no-zoom` class that we can add to images where we _aren't_ specifying them in Markdown. I could have just added an `<em>` wrapper in React, but the classname seemed cleaner in that context.

The selector here essentially says, select an image for which _neither_ of the above are true. That is, if it's wrapped in emphasis _or_ has the `.no-zoom` class, the selector won't apply.

eak12913 (approved):
LGTM!

PR #148, opened 2022-01-20
Sidebar index for our guides

ebeneliason:


ebeneliason:
I think this is finally ready for review. I've squashed it down into two commits. The first introduces the new guides sidebar and index pages, adds back buttons to guides, and cleans up some styling. The second shuffles the RefArch guide down one level (so it reflects the patterns of other guide categories), updates its naming, and adds an admonition to clarify that it's just an example.

eak12913 (dismissed):
I've spot checked as much as I could and don't see any glaring issues. The only comment I'll make is that the style guides end up without any sidebar and therefore there's no "back" button from them. They are "flatter" than the other guides - so I'm pretty sure that the issue here is that there's nowhere to put a "back" button when there's no sidebar - and no sidebar is needed on those pages.

I guess my only comment is that perhaps we'll need another back button solution for pages without a sidebar? Or perhaps we have a sidebar of exactly two items: the back button and the category index page?

oredavids (approved):
Reviewed live.

PR #149, opened 2022-01-24
Use published docs-soucer package

oredavids:


oredavids:
Eugene and I discussed the pros and cons of switching to NPM and concluded that we will stick to using the current approach of the GitHub link because using NPM introduces another access management layer and we already have Github. Switching will also mean we have to handle secrets earlier than we intend to.

PR #151, opened 2022-01-24
WIP: Production framework ingredients cards

ebeneliason:
@brikis98 FYI, here's a quick example of what the Gruntwork Production Framework ingredients could look like presented as cards. We'd want to refine the icons — this is FPO to get a sense of how it could look.

[Direct link](https://deploy-preview-151--pensive-meitner-faaeee.netlify.app/docs/guides/production-framework/#the-elements-of-the-gruntwork-production-framework)


brikis98:
From a look & feel perspective, the cards are nice. As you said, the icons need a bit of tweaking—we'd want them to be consistent with the image—but that's a minor detail.

However, from a readability perspective, I'm not sure we want clickable links there. It seems to indicate the user should click on them immediately, right at the top of that post, without reading any of the other context on what the framework is, who it's for, the pre-requisites, etc. 

So perhaps we keep the original (non-clickable) image at the top—and also make it the preview image when sharing this thing on social media—and put the cards at the bottom of this intro post? Because at the bottom, we very much want people to click on those links.

ebeneliason:
>However, from a readability perspective, I'm not sure we want clickable links there. It seems to indicate the user should click on them immediately…

We have options. The `Card` component in use here supports an `href` property, but if you omit it then the cards aren't hoverable/clickable. In other words, we can make it "for show" only, if we want.

Since social media sharing seems to be the top motivator for an image up front, we could also address that head on. It's possible to add meta tags to our pages, so perhaps [we should provide an OpenGraph image and other metadata](https://css-tricks.com/essential-meta-tags-social-media/) to ensure that works exactly as desired regardless of whether the image is shown in the page.

brikis98:
> > However, from a readability perspective, I'm not sure we want clickable links there. It seems to indicate the user should click on them immediately…
> 
> We have options. The `Card` component in use here supports an `href` property, but if you omit it then the cards aren't hoverable/clickable. In other words, we can make it "for show" only, if we want.
> 
> Since social media sharing seems to be the top motivator for an image up front, we could also address that head on. It's possible to add meta tags to our pages, so perhaps [we should provide an OpenGraph image and other metadata](https://css-tricks.com/essential-meta-tags-social-media/) to ensure that works exactly as desired regardless of whether the image is shown in the page.

Using meta tags seems like a good idea for social media.

If the cards aren't clickable, is there a reason you prefer them over the image? Clickable cards at the end of the post, where there's already a mini outline that we want to be clicked, makes sense. But is there an advantage of cards over the image at the top? I don't have a strong attachment one way or the other, just curious.

ebeneliason:
> If the cards aren't clickable, is there a reason you prefer them over the image?

I think it's a combination of factors.
* It's more consistent with chapter outlines in our other guides, and I like consistency.
* It feels a bit more aesthetically integrated, and works for both light and dark themes.
* It's searchable/accessible/selectable — I try to avoid images that are mostly text, when possible.
* It scales better — when scaling up while zooming (for accessibility) it doesn't get blurry; when scaling down to mobile, it can wrap instead of getting really small.  (On that note, this image is already blurry at default size — if we keep it, we should create a higher res asset.)

> Clickable cards at the end of the post, where there's already a mini outline that we want to be clicked, makes sense.

I'm not so worried about offering clickable cards up front, myself. If users _want_ to jump into the deep end on a given topic, perhaps we should let them? The back button and sidebar can ensure they don't get lost.

One more note — that outline at the end is slightly different, since it covers the top level chapters and their subsections, whereas the one up front is only listing the ingredients, which amount to the subsections of the first chapter.

brikis98:
> > If the cards aren't clickable, is there a reason you prefer them over the image?
> 
> I think it's a combination of factors.
> 
> * It's more consistent with chapter outlines in our other guides, and I like consistency.
> * It feels a bit more aesthetically integrated, and works for both light and dark themes.
> * It's searchable/accessible/selectable — I try to avoid images that are mostly text, when possible.
> * It scales better — when scaling up while zooming (for accessibility) it doesn't get blurry; when scaling down to mobile, it can wrap instead of getting really small.  (On that note, this image is already blurry at default size — if we keep it, we should create a higher res asset.)
> 

These are fairly convincing arguments!

> > Clickable cards at the end of the post, where there's already a mini outline that we want to be clicked, makes sense.
> 
> I'm not so worried about offering clickable cards up front, myself. If users _want_ to jump into the deep end on a given topic, perhaps we should let them? The back button and sidebar can ensure they don't get lost.

Having looked back at the post, I must admit I'm still worried this is going to lead users astray:

Screen Shot 2022-01-27 at 10 12 18 AM

If you read the text, it basically says, "In this guide, we are going to share our opinionated, step-by-step framework," and just below that, it says "The Elements of the Gruntwork Production Framework," with 5 links to click. You can't see that there's any further text below that, so it's not obvious there's a ton more intro text, and I'm guessing the vast majority of people will click on one of those 5 links. That won't be a terrible experience, but they'll miss all the context in the intro, and they won't know to click "back," as it's not obvious there was anything more in the intro in the first place!

> 
> One more note — that outline at the end is slightly different, since it covers the top level chapters and their subsections, whereas the one up front is only listing the ingredients, which amount to the subsections of the first chapter.

micahch (approved):

ebeneliason:
> If you read the text, it basically says, "In this guide, we are going to share our opinionated, step-by-step framework," and just below that, it says "The Elements of the Gruntwork Production Framework," with 5 links to click. You can't see that there's any further text below that, so it's not obvious there's a ton more intro text, and I'm guessing the vast majority of people will click on one of those 5 links.

Yeah, I think that's a valid concern. I'm interested in some exploring visual refinements here that could mitigate the issue, but  also concede that they don't _need_ to be clickable, and perhaps we can avoid misdirection if they aren't. As for the visual tweaks, I'd like to play with:

* Adjust padding to tighten things up a bit.
* Use a "stacked" treatment where the cards all feel like one combined unit, with "creases" separating them instead of gaps.

If I find any spare time this week I'll play with it. Do you happen to have some SVG sources for the icons?

brikis98:
I made the diagram in draw.io: https://github.com/gruntwork-io/docs/blob/master/static/img/guides/production-framework/gruntwork-production-framework.drawio

eak12913:
Closing this PR as it is currently stale. I will save the branch and we can resurrect this if we want to do something with it later.

PR #154, opened 2022-01-25
[APT-1691] Add Contributing and Forking sections

eak12913:
This PR adds two new sections to the Guides area:
- Contributing
- Forking

These new pages will live under a new "working-with-our-code" area. Additionally we're moving the "versioning" guide from "stay-up-to-date" to "working-with-our-code".


_docs-sources/guides/working-with-code/forking.md
```diff
@@ -0,0 +1,56 @@
...
+While forking is allowed under the Gruntwork Terms of Services, it has some downsides:
+
+- You have to do a lot of work up-front to copy the repos, releases, and pre-compiled binaries and update internal
+  links.
+- You have to do more work to run this process on a regular basis and deal with merge conflicts.
+- If your team isn’t directly using the Gruntwork GitHub repos on a regular basis, then you’re less likely to
+  participate in issues and pull requests, and you won’t be benefiting as much from the Gruntwork community.
+
+So, whenever possible, use the code directly from the `gruntwork-io` GitHub org, as documented in
+[Using Terraform Modules](#using_terraform_modules) and [Using scripts and binaries](#using_scripts_binaries). If your team relies on NPM, Docker Hub, Maven Central,
```
ebeneliason, line 54: Same comment as above regarding anchor links.

_docs-sources/guides/working-with-code/forking.md
```diff
@@ -0,0 +1,56 @@
...
+
+You’ll want to automate the entire process above and run it on a regular schedule (e.g., daily). The Gruntwork
+Infrastructure as Code Library is updated continuously, both by the Gruntwork team and contributions from our community
+of customers (see the monthly [Gruntwork Newsletter](https://blog.gruntwork.io/tagged/gruntwork-newsletter) for details),
+so you’ll want to pull in these updates as quickly as you can.
+
+## How to use your forked code
+
+Once you’ve forked the code, using it in is very similar to what is outlined in [Using Terraform Modules](#using_terraform_modules) and
+[Using scripts and binaries](#using_scripts_binaries), except for the following differences:
```
ebeneliason, line 38: These anchor links are broken. I think the first could possible go to [this intro section](https://deploy-preview-154--pensive-meitner-faaeee.netlify.app/docs/intro/first-deployment/using-terraform-modules), but we don't have a good answer for the latter at the moment. Thoughts?

eak12913: I wonder if the right thing to do is to allow it to be broken. My thought process:
- If we don't have a place to link it to and we link it to an existing place which doesn't really help the user understand anything then we'll have no way in the future to discover this (unless someone tells us about it)
- If we leave it broken - or update the link to a known broken thing (eg: "replace-me-when-we-have-this-section") then we'll be able to punt dealing with the problem until we want to turn on the link validator which we are about to build. At that time - perhaps we'll have a better section for this?

ebeneliason: Where would you propose we link to? Two candidates for this content seem to be
- Somewhere in the intro guide
- The forthcoming Tools reference section

It's weird to leave it broken, though. If we link to tools, perhaps we should have a stub there with an admonition saying this isn't done, and with a link back to the original section of the guide on Gruntwork.io? Or even a redirect? That would feel better, I think.

ebeneliason (approved):
Looks good!

PR #155, opened 2022-01-26
[WIP][APT-1673] Update regenerate scripts

oredavids:
#### What does this PR do?
- Updates the docs site to use new **docs-sourcer** features

**NOTE**: This has a WIP label because the docs-sourcer package reference needs
 to be updated once the [corresponding PR on the docs-sourcer repo](https://github.com/gruntwork-io/docs-sourcer/pull/8) is merged.
 
[APT-1673]

[APT-1673]: https://gruntwork.atlassian.net/browse/APT-1673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,9 @@
     "write-translations": "docusaurus write-translations",
     "write-heading-ids": "docusaurus write-heading-ids",
     "typecheck": "tsc",
-    "regenerate:local": "docs-sourcer"
+    "regenerate": "env-cmd docs-sourcer",
+    "regenerate:local": "yarn regenerate --local",
+    "regenerate:remote": "yarn regenerate --remote"
   },
   "dependencies": {
     "@docusaurus/core": "^2.0.0-beta.14",
@@ -22,6 +24,7 @@
     "@svgr/webpack": "^5.5.0",
     "clsx": "^1.1.1",
     "config": "^3.3.6",
+    "env-cmd": "^10.1.0",
     "file-loader": "^6.2.0",
     "plugin-image-zoom": "ataft/plugin-image-zoom",
     "posthog-docusaurus": "^1.0.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2936,11 +2936,6 @@
   resolved "https://registry.yarnpkg.com/@types/btoa-lite/-/btoa-lite-1.0.0.tgz#e190a5a548e0b348adb0df9ac7fa5f1151c7cca4"
   integrity sha512-wJsiX1tosQ+J5+bY5LrSahHxr2wT+uME5UDwdN1kg4frt40euqA+wzECkmq4t5QbveHiJepfdThgQrPw6KiSlg==
 
-"@types/config@^0.0.39":
-  version "0.0.39"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -13,7 +13,9 @@
     "write-translations": "docusaurus write-translations",
     "write-heading-ids": "docusaurus write-heading-ids",
     "typecheck": "tsc",
-    "regenerate:local": "docs-sourcer"
+    "regenerate": "env-cmd docs-sourcer",
+    "regenerate:local": "yarn regenerate --local",
+    "regenerate:remote": "yarn regenerate --remote"
```
ebeneliason, line 18: Won't these two lines need to update in accordance with the `docs-sourcer` parameter adjustment? Something like…
```suggestion
    "regenerate:local": "yarn regenerate --plugin local",
    "regenerate:remote": "yarn regenerate --all"
```

ebeneliason: Or: maybe there's only…
```suggestion
    "regenerate": "env-cmd docs-sourcer --all",
    "regenerate:local": "env-cmd docs-sourcer --plugin local"
```

ebeneliason: Forgive me, Ore, as I propose yet a third option. I'll add a little more context this time so you can follow my reasoning. While this all boils down to choosing the appropriate switches and defaults, that feels important to me because which we choose help to suggest the expected pattern of use, which makes it a UX concern.

In a standalone context, the tool is responsible for running a set of plugins in order to generate a docs site from various sources. The same can be said when running the tool in our cloud service — it's concerned with _everything_. For this reason, I think I may have misled in suggesting the need for a `--all` in the first place. That should probably be the default _for the `docs-sourcer` itself_.

In the context of the docs site, where we leverage yarn, it's _more_ likely that a user will want to do local-only regeneration, to view content they just edited. That's even more true in the long run, when the cloud process will manage changes to remote content automatically. There are still use cases for running remote generation in the docs repo manually, but they seem secondary.

For these reasons, I would propose that the `docs-sourcer` command runs all plugins by default, and accepts the `--plugin` and `--repo` arguments as a way to filter that to specific plugins. I'd then propose one of two options for the docs site itself

**OPTION 1:** Closest to standard usage, with just a shortcut for local generation as a convenience:
```
    "regenerate": "env-cmd docs-sourcer",
    "regenerate:local": "yarn regenerate --plugin local"
```

**OPTION 2:** Emphasizing the appropriate default for the context, at the risk of essentially reversing the default:
```
    "regenerate": "env-cmd docs-sourcer --plugin local",
    "regenerate:all": "env-cmd docs-sourcer"
```

Thoughts?

oredavids: I am leaning with option 2 in the context of the docs repo. It is trivial to change so we test out any and switch if needed.

@eak12913 thoughts?

ebeneliason: Upon reflection, I like option 2 best also. I think it reflects the pattern of use better, and no one but us will care that it follows a different pattern than the underlying `docs-sourcer` tool itself.

oredavids:
Closing in favour #156

PR #159, opened 2022-01-30
Fix AWS region for `docker login` command

jalessio:
I think the `docker login` command should be targeting `us-west-2` since that is the region where the ECR repo is located.


rhoboat:
I've got a [PR to change this upstream](https://github.com/gruntwork-io/terraform-aws-architecture-catalog/pull/562/files). We'd have to re-pull the generated docs there (or from the for-production examples in the service-catalog repo).

eak12913:
@jalessio - Thanks for the submission. I'm going to close this PR in favor of #162 as the file you've changed (though a logical thing to change) is actually a generated file. We haven't done a great job yet with exposing an easy way to submit a change that gets properly built. Thank you for the contribution anyway!

PR #164, opened 2022-01-31
[APT-1692] Strip /docs/ prefix from all site URLs

ebeneliason:
This takes advantage of the Docusaurus base page parameter to avoid the redundant "/docs/…" bit of the path for every file on the docs site. All internal links are updated accordingly.

### DIFF
--- a/_docs-sources/guides/build-it-yourself/achieve-compliance/deployment-walkthrough/deploy-landing-zone-solution/apply-account-baseline-root-to-root-account.md
+++ b/_docs-sources/guides/build-it-yourself/achieve-compliance/deployment-walkthrough/deploy-landing-zone-solution/apply-account-baseline-root-to-root-account.md
@@ -58,9 +58,9 @@ We’ll be using the `landingzone/account-baseline-root` module from [terraform-
 :::info
 
 This guide will use [Terragrunt](https://github.com/gruntwork-io/terragrunt) and its associated file and folder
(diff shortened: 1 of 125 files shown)


brikis98:
Doesn't this break all existing URLs? E.g., I already shared https://docs.gruntwork.io/docs/guides/production-framework/ on social media... Can we add a redirect from `/docs` to `/`?

eak12913:
I believe we can add redirects according to: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects

ebeneliason:
Yeah, the `createRedirects` function looks like it would do the trick. I had thought we caught this matter early enough to not have much impact for SEO or UX, but we should go ahead and implement the redirect before we release the change. That utility will no doubt be useful in the future, too. Thanks for the callout.

ebeneliason:
Oh, but this is just a client site redirect. Ideally we'd do this server side with a proper 302. We should look into configuring this on our server instead of in Docusaurus first. @eak12913 can you explore what's possible in your config?

eak12913:
We've discussed this internally and we're not going to release this as-is. Instead we're going to merge this PR and then we're going to update our S3 bucket to have a routing rule according to: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-websiteconfiguration-routingrules.html

We've already applied this rule to stage and will now test what happens after this PR gets merged to make sure there are no issue. If we're successful then we'll make a similar change to `prod` ahead of releasing this.

eak12913 (approved):
LGTM - merging to do a final test on stage.

eak12913:
@brikis98: Update - this has been confirmed as working in stage such that: https://docs.dogfood-stage.com/docs/guides/production-framework/ will receive a 301 and redirect to the non- `docs/` path.

PR #165, opened 2022-02-01
Sidebar generator

ebeneliason:
Adds a new script, run via `yarn sidebar`, capable of generating sidebar.js files for any direcory (or directories) in `_docs-sources` (or in the output `docs` directory). It includes support for the category index convention, and supports adding an optional back button. Tests included.


eak12913 (commented):
This looks great! The only thing we should do is to add a small section to the main README to point this new feature out.

ebeneliason:
> The only thing we should do is to add a small section to the main README to point this new feature out.

Indeed! That's in the AC, in fact, but I forgot to include it. I'll push another commit, hopefully later today.

ebeneliason:
I've pushed a new README section (and some other minor changes to it), as well as a few additions and fixes in the tests.

eak12913 (approved):
LGTM!

scripts/sidebar-lib.js
```diff
@@ -73,9 +73,9 @@ async function generateSidebar(dir, opts = {}) {
 
   for (const fullPath of fullPaths) {
     if (fs.statSync(fullPath).isDirectory()) {
-      // process subdirectories recusively (back button is top level only!)
```
ebeneliason: I had already called this out explicitly, and yet somehow forgot to omit it in the opts the first time around.

eak12913 (approved):
LGTM!

PR #167, opened 2022-02-04
Consolidate duplicated instructions on AMI building

yorinasub17:
The instructions for building AMIs with packer was duplicated in two places, so this consolidates it into one.


eak12913:
I think the build is failing because you've deleted a page - but that page is still being referenced in the sidebar. Or there's some link on some other page that's referencing the page you deleted.

eak12913 (approved):
LGTM! Thank you!

PR #168, opened 2022-02-04
[APT-1657] Add build identifier meta tags

oredavids:
[APT-1657]

Screen Shot 2022-02-04 at 10 45 41 AM

[APT-1657]: https://gruntwork.atlassian.net/browse/APT-1657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ebeneliason (approved):
Reviewed live. We discussed adding a generic key rather than using `CIRCLE_SHA1` directly, but we don't need to hold this up based on that.

PR #172, opened 2022-02-04
Undo an accidental UI change to the production framework guide

ebeneliason:
This slipped in with a much larger change, so we have to undo it
rather than reverting the commit.


eak12913 (approved):
LGTM!

oredavids (approved):
LGTM.

PR #193, opened 2022-02-15
[APT-1700] Fix relative links

eak12913:
This PR addresses an issue where some relative links were resulting in 404s if users landed directly on those pages rather than having navigated to them via other links in the application.


ebeneliason (commented):
`docs/guides/reference-architecture/configuration-guide/index.md`

This file only appears with changes in `docs/`. It appears the `docs-sourcer` did have cause to run here, given that a file was absent in `_docs-sources`. However, this commit has many files. Perhaps when it bails due to the number of files it doesn't skip the commit as it should?

ebeneliason (approved):
This looks good.

docs/guides/stay-up-to-date/terraform/terraform-15/deployment-walkthrough/step-1-update-your-code-to-be-compatible-with-terraform-0-14.md
```diff
@@ -10,19 +10,19 @@ If you haven’t already, you need to:
...
-        [Terraform 0.12 upgrade guide](../../terraform-12).
+        [Terraform 0.12 upgrade guide](../../terraform-12/index.md).
 
     2.  If you’re still on Terraform 0.12, see our
-        [Terraform 0.13 upgrade guide](../../terraform-13).
+        [Terraform 0.13 upgrade guide](../../terraform-13/index.md).
 
     3.  If you’re still on Terraform 0.13, see our
-        [Terraform 0.14 upgrade guide](../../terraform-14).
+        [Terraform 0.14 upgrade guide](../../terraform-14/index.md).
```
ebeneliason, line 19: Do we need the `/index.md` on these? That sort of eliminates the advantage of using the index convention.

eak12913: We discussed this live - but for posterity: Yes - we do need this otherwise we get the 404 when the site is hosted in S3. Notably, the final URL (rendered into HTML) will not have `/index.md` in it.

PR #194, opened 2022-02-17
[APT-1732] Update version badge to support last modified version

ebeneliason:


ebeneliason:
@eak12913 I've updated this PR with the changes to the service API pages that leverage the new docs-sourcer change in [this PR](https://github.com/gruntwork-io/docs-sourcer/pull/24). However, we'll still need to bump the version after releasing in that repo, as usual.

Also, there's a gotcha here that we may need to address. The "last modified version" reported here does not always match the first version displayed when clicking on the release notes button, because I'm currently exact-matching the release notes on e.g. "data-stores/aurora", whereas that same filter string in GitHub does a less strict match, which (I'm guessing) matches on the full string _or_ its subcomponents ("data-stores", "auroroa"). It's possibly it's doing something even fuzzier but I can't be sure.

I'm not sure which is more correct. It does seem that occasionally the release notes will just list the category (e.g. "data-stores") if a change implicates the whole collection rather than a specific instance. As such, my inclination is to change my matching logic in the `docs-sourcer` to perform the match against any of ["data-stores/aurora", "data-stores", "aurora"], which I _think_ will closely-if-not-exactly result in version numbers that match the release notes filter. I have that change ready to go on my end but haven't had a chance to test it yet.

eak12913 (approved):
LGTM. We're aware of the issues with TLS-scripts and will address that in follow up PRs

PR #199, opened 2022-02-18
Update macie settings for achieve compliance guide

yorinasub17:
This updates a bug in the macie settings for the achieve compliance guide, where it uses settings that don't make sense (creating a macie bucket in logs and security when they are rolling up to `root` account).


ebeneliason (approved):
LGTM.

PR #216, opened 2022-03-01
[WIP] Expanded intro to include onboarding content

pete0emerson:
This content mirrors content that we currently do when onboarding customers live.

Consider this first draft. Looking for comments, and I will also come back to this soon with a fresh pair of eyes.

### DIFF
--- a/_docs-sources/intro/overview/intro-to-gruntwork.md
+++ b/_docs-sources/intro/overview/intro-to-gruntwork.md
@@ -15,3 +15,65 @@ Gruntwork products focus on Amazon Web Services (AWS). Support for other public 
 ### Gruntwork uses Terraform
 
 All Gruntwork products are built on and fully compatible with [open source Terraform](https://terraform.io). The one exception to this is the [Gruntwork Reference Architecture](https://gruntwork.io/reference-architecture/), which uses [Terragrunt](https://terragrunt.gruntwork.io/) (one of our open source tools) to implement an end-to-end architecture.
+
+### Core modules at our core
+
+In each of our repositories, you will find the following structure:
+
+```
+├── README.md
+├── examples/
+├── modules/
+├── test/
+```
+
+The `README.md` file gives an overview of the modules inside the repository, and may include diagrams and information on how to use the modules.
+
+The `examples/` directory contains small bits of code that launch the module. You can use these examples to craft your own infrastructure.
+
+The `modules/` directory contains the code for each of the modules. This is where the meat of the work is done.
+
+The `test/` directory contains all of our tests that go through our CI/CD pipeline for every commit. It uses our open source [Terratest](https://terratest.gruntwork.io/) Go library. The tests leverage the examples in the `examples/` directory.
+### Services bring multiple modules together
+
+One level up from our modules is our [Service Catalog](https://docs.gruntwork.io/guides/production-framework/ingredients/service-catalog/). The Service Catalog brings modules that are often used together into their own module that you can instantiate.
+
+One difference in the [file layout for our Service Catalog](https://github.com/gruntwork-io/terraform-aws-service-catalog/) can be seen in the `examples/` folder:
+
+```
+├── examples
+│   ├── for-learning-and-testing
+│   │   ├── data-stores
+│   │   ├── landingzone
+│   │   ├── mgmt
+│   │   ├── networking
+│   │   └── services
+│   └── for-production
+│       └── infrastructure-live
+```
+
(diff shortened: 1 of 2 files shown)


_docs-sources/intro/overview/intro-to-gruntwork.md
```diff
@@ -15,3 +15,65 @@ Gruntwork products focus on Amazon Web Services (AWS). Support for other public
 ### Gruntwork uses Terraform
 
 All Gruntwork products are built on and fully compatible with [open source Terraform](https://terraform.io). The one exception to this is the [Gruntwork Reference Architecture](https://gruntwork.io/reference-architecture/), which uses [Terragrunt](https://terragrunt.gruntwork.io/) (one of our open source tools) to implement an end-to-end architecture.
+
+### Core modules at our core
```
eak12913, line 19: Consider rephrasing. Perhaps "General anatomy of a Gruntwork module"?

_docs-sources/intro/overview/intro-to-gruntwork.md
```diff
@@ -15,3 +15,65 @@ Gruntwork products focus on Amazon Web Services (AWS). Support for other public
...
+```
+
+The `README.md` file gives an overview of the modules inside the repository, and may include diagrams and information on how to use the modules.
+
+The `examples/` directory contains small bits of code that launch the module. You can use these examples to craft your own infrastructure.
+
+The `modules/` directory contains the code for each of the modules. This is where the meat of the work is done.
+
+The `test/` directory contains all of our tests that go through our CI/CD pipeline for every commit. It uses our open source [Terratest](https://terratest.gruntwork.io/) Go library. The tests leverage the examples in the `examples/` directory.
+### Services bring multiple modules together
```
eak12913, line 37:
```suggestion

### Services bring multiple modules together
```

_docs-sources/intro/overview/intro-to-gruntwork.md
pete0emerson: You don't like my "Core ... core" humor? /sigh

I like your phrasing. Committed in f779c49

ebeneliason:
Hey Pete, thanks for this! The content makes sense, but I'm not sure the very first page of the intro guide is quite the right place for it. It seems to jump the gun a little, as the guide is structured first with a high level overview of what Gruntwork is, how it works, core concepts, gaining access to the code, and setting up the environment before actually getting into the details of the code itself. I think this new content would belong somewhere after all that. Perhaps we should add a new first page to the "Deploy Your First Module" section to cover these technical basics?

I'm also noting some similarities in the content to the guides that now live in the [introduction to the service catalog](https://docs.gruntwork.io/reference/services/intro/overview). That's not a problem — it likely still makes sense to cover the basics somewhere in the intro itself so users have some grounding.

pete0emerson:
Ulimately what I want to create is the written form of the current onboarding experience. I want that easily discoverable, as well. I want people who skip our live onboarding experience to have a decent written replacement.

Perhaps this goes in it's own "Onboarding" section under "Overview" ? or maybe in "Core Concepts"?

The content in [into to the service catalog](https://docs.gruntwork.io/reference/services/intro/overview) definitely has a bunch of overlap, but I wouldn't replace our onboarding experience with it.

Document organization is hard! I'm not sure what the right solution is, and am definitely open to suggestions.

ebeneliason:
> The content in [into to the service catalog](https://docs.gruntwork.io/reference/services/intro/overview) definitely has a bunch of overlap, but I wouldn't replace our onboarding experience with it.

I definitely agree with that. It's important to put info in front of people when they need it. If the onboarding materials don't need to be as in depth, we could also cover the basics and then link to more comprehensive docs elsewhere.

> Perhaps this goes in it's own "Onboarding" section under "Overview" ? or maybe in "Core Concepts"?
> Document organization is hard! I'm not sure what the right solution is, and am definitely open to suggestions.

I think your aperture may be a bit narrow. From what I hear, you're looking to create a single page somewhere that covers all the same material you cover in an onboarding call. However, I'd encourage thinking of this entire intro guide in that way. I still think it makes more sense to cover the ins and outs of the code structure _after_ the user has gotten access to the code, which means (in the current structure) that would go in "Deploy Your Module."

Perhaps an new **Overview** section could begin with e.g. "Before deploying your first module, it's important to understand the general structure of a Gruntwork module…" Notably, that whole section of the guide is still built around the module catalog, and might need to be rewritten to focus on service modules instead. In fact, it may even make sense to swap the order in which you present these ideas in the bit you've written here, talking about the higher level services first before calling attention to the lower level building blocks you can leverage if you need more control.

pete0emerson:
Closing PR.

PR #218, opened 2022-03-07
Run regenerate local command

oredavids:
#### What does this PR do?
- Fixes the premature merge of https://github.com/gruntwork-io/docs/pull/217


ebeneliason (approved):
Ship it.

ebeneliason:
@eak12913 FYI — this change prompted the `docs-sourcer` to run, even though it produced an empty commit. I suppose that's expected since this fix is exactly what it would have produced itself. Stepping one step further, [this PR](https://github.com/gruntwork-io/docs/pull/217) made a chance in `_docs-sources/` but _didn't_ update the output docs. Should we consider catching that case with our automation?

PR #233, opened 2022-03-24
Updated to use yml and apply iam group for mfa

scotlyt:
I updated it to use the users.yml reference instead of the old .hcl version.

I also added the group iam-user-self-mgmt to any user that doesn't have full access because they will not be able to update their MFA within the security account with only account specific permissions.

I am also suggesting `Remove temporary login lines` so those old default initial passwords don't keep on showing up in the terraform output whenever you are adding in more users.


eak12913:
@scotlyt Thanks so much for submitting this PR! I've ported your exact changes into #234 solely because we have an automated process that runs, but that process is not capable of operating on your forked version of the `docs` repo. 

Thanks again for the help!

scotlyt:
great thanks

PR #234, opened 2022-03-24
User submitted PR for #233

eak12913:
This PR is direct commit version of #233. The reason we're creating this is because the original PR is based on a forked repo and we can't currently run our `docs-sourcer` process in the user's fork.


yorinasub17 (approved):
Content updates LGTM! However, is this going to get overwritten when we update architecture catalog? We might need to reflect this in the architecture catalog first...

eak12913:
> is this going to get overwritten when we update architecture catalog? We might need to reflect this in the architecture catalog first...

I don't believe that this section has been automated yet. So for the moment - I think it's safe to merge and ship this. How can we also get this reflected in the ref arch docs?

PR #284, opened 2022-04-15
Add missing link to production framework

brikis98:
The production framework guide promises a closer look at app & infra pipelines later in the guide... But there is no such content later. Therefore, I've replaced that promise with a link to a blog post we wrote before that has exactly this content.


infraredgirl (approved):

brikis98:
Thanks for the review!

PR #287, opened 2022-04-15
Remove urgent-support email address and change support URL

pete0emerson:
We added this email address a little while ago, but changed tack. This removes it, and also updates the Zendesk URL to be one that is on the gruntwork.io domain.


josh-padnick (approved):
LGTM. I think there are additional optimizations we can make on this page, but we'll leave those for separate PRs!

PR #311, opened 2022-04-26
Include bash style guide

rhoboat:
I'd like to just start here, with this external link.

Follow-ups:
- [ ] Create our extended version of google's guide.
    - Description for this issue: We have only one extra thing to extend this base guide with so far, and that's using `local -r` instead of `local readonly`. We can add more rules as we go along,


rhoboat:
@ebeneliason @eak12913 @oredavids I'm not sure it reads so clearly that on this page, we say "learn how Gruntwork formats our X" because... aren't _we_ Gruntwork?

ebeneliason (approved):
Looks good to me.

eak12913:
It looks fine to me Rho! I'm going to merge it.

PR #332, opened 2022-05-13
CircleCI now supports static IPs

pete0emerson:


pete0emerson:
Superceded by #333 (via the docs site)

PR #406, opened 2022-06-22
Update terraform-style-guide.md to recommend blocks before used.

rhoboat:
[Slack thread for context](https://gruntwork-io.slack.com/archives/C01M9MXBNTF/p1655829099362759). The PR is meant to act as a way to vote 👍 or 👎 making this change. I am not strongly opinionated on it, myself!

### DIFF
--- a/docs/guides/style/terraform-style-guide.md
+++ b/docs/guides/style/terraform-style-guide.md
@@ -131,7 +131,7 @@ There is no standard on ordering the sections, but as a rule of thumb the follow
 - Provider blocks, if needed.
 - The main component of the module (e.g., the `aws_ecs_service` resource for the ECS service module).
 - All other sections.
-- Any `data` blocks (at the bottom).
+- `locals` and `data` blocks just before where they're used.
 
 #### Conditionals
 
--- a/docs/guides/style/terraform-style-guide.md
+++ b/docs/guides/style/terraform-style-guide.md
@@ -397,6 +397,6 @@ func TestECS(t *testing.T) {
 <!-- ##DOCS-SOURCER-START
 {
   "sourcePlugin": "local-copier",
-  "hash": "ebf33215ca11bf62867a076882927304"
+  "hash": "3612e029d161a92b64b4e9b459901244"
 }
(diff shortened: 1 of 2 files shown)


rhoboat:
faaeee!

docs/guides/style/terraform-style-guide.md
```diff
@@ -131,7 +131,7 @@ There is no standard on ordering the sections, but as a rule of thumb the follow
 - Provider blocks, if needed.
 - The main component of the module (e.g., the `aws_ecs_service` resource for the ECS service module).
 - All other sections.
-- Any `data` blocks (at the bottom).
+- `locals` and `data` blocks just before where they're used.
```
iangrunt, line 134: I think it depends on both cases. I tend to keep both blocks on the top and consolidate as much into one `local {}` as possible, if they're producing values that are used in more than one resource/module call. For an AMI lookup or a very specific string calculation that requires a `local` to be used, those are cases I think putting them directly above what is going to use them next helps the reader.

yorinasub17: One thought: we used to separate out all data sources to a separate `dependencies.tf` file so that users can scan what resources we expect to exist for the module and are looked up (this pattern is employed to some extent in `terraform-aws-eks` repo). Similarly, one of the contractors pointed out that it might be better for `locals` to be separated out into a separate `locals.tf` file to make it easier to scan, search, and find.

I personally think that might be a better practice than having everything in one giant `main.tf` file at the top. Thoughts on that approach as an alternative?

rhoboat: I'm okay with either way. I find it easier to scan and search if they're in the same file, but I could see myself creating a plugin or shortcut that opens all *.tf files automagically into separate window-splits. I tend not to remember things when flipping between files that aren't visible at the same time.

For long term, separate files would be cleaner.

eak12913:
It seems like we've not taken any action here. I am closing this PR and will update the team to ask. Folks can re-open as necessary if the changes in here still need to be made.

PR #433, opened 2022-07-02
Update import-existing-resources-from-the-root-account-into-terraform…

rstewart-sms:
…-state.md

When using terragrun aws-provide-patch, the strings passed using --terragrunt-override-attr should be wrapped in single quotes.  If the strings are not surrounded by single quotes then running the command results in the error "Could not determine underlying type of JSON string".


eak12913:
We've got an internal process that needs to run for these docs changes. So I've opened up https://github.com/gruntwork-io/docs/pull/443. It's got your original commit (so you should still get credit for the changes) and it also has a new commit from our bot to add in the missing auto-generated stuff. I'll close this PR and will get that one merged.

eak12913:
Closing this PR as we've now merged #443

PR #443, opened 2022-07-13
Update import-existing-resources-from-the-root-account-into-terraform…

eak12913:
…-state.md

When using terragrun aws-provide-patch, the strings passed using --terragrunt-override-attr should be wrapped in single quotes.  If the strings are not surrounded by single quotes then running the command results in the error "Could not determine underlying type of JSON string".


yorinasub17 (approved):
LGTM!

PR #444, opened 2022-07-15
how-it-works.md: Remove mention of Gruntwork Professional Services

iangrunt:
We no longer offer this service.


yorinasub17 (approved):
LGTM!

PR #553, opened 2022-10-08
Subscriber notices now work in more cases

ebeneliason:
Subscriber notices used to work for basic links, but not in cases where links wrapped components, such as the Cards we use frequently to organize content. In addition to ensuring notices appear in more cases, this fixes the logic for extracting the repo name to ensure it appears in the notice.


eak12913 (approved):
LGTM!

PR #565, opened 2022-10-25
Add warning about forking our code

ebeneliason:
Forks of private Gruntwork repos will be deleted if their creator loses access, which can be both unexpected and highly disruptive, so a warning message is now shown.


ebeneliason:
Whoops — opened this one directly against the source file, which means the automation didn't trigger to to update the output (it only works the other direction).

PR #595, opened 2022-12-01
Update releases through Nov

brikis98:
Re-run releases plugin in `docs-sourcer`.

Direct link to preview: https://deploy-preview-595--pensive-meitner-faaeee.netlify.app/guides/stay-up-to-date/

**Update**: based on PR feedback in https://github.com/gruntwork-io/docs-sourcer/pull/93, I switched the `docs-sourcer` to a different HTML escape function, so now far more characters get escaped. This is good... but it means almost every single release file changed, so this PR is now rather large.

### DIFF
--- a/docs/guides/stay-up-to-date/index.md
+++ b/docs/guides/stay-up-to-date/index.md
@@ -16,6 +16,7 @@ import CardGroup from "/src/components/CardGroup"
 
 <CardGroup cols={1} gap="1rem" stacked equalHeightRows={false} commonCardProps={{padding: "1.25rem"}}>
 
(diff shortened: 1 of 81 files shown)


brikis98:
OK, fixed date handling to be consistent across runs in https://github.com/gruntwork-io/docs-sourcer/pull/93/commits/6fec5764aa167e0ed9a5e4fd544d74ff7562e957. Note that a couple dates are still different, but it should be consistent going forward from here due to that fix.

eak12913 (approved):
LGTM!

PR #598, opened 2022-12-08
CORE-26 - Add Ref Arch Prerequisites Guide

pete0emerson:
This is my first cut at adding the Ref Arch Prerequisites Guide per https://gruntwork.atlassian.net/browse/CORE-26


bethadele:
This is looking great. :shipit:

PR #605, opened 2023-01-17
Upgrade to docusaurus v2.2.0

oredavids:
#### Sanity Check
- [x] Docusaurus now requires node v16.14.0; do we need to make infra changes for stage & prod?


docusaurus.config.js
```diff
@@ -243,11 +243,6 @@ const config = {
         defaultMode: "light",
         disableSwitch: false,
         respectPrefersColorScheme: false,
-        /* we clear these icons out and use CSS to swap in our own images instead */
-        switchConfig: {
```
eak12913, line 247: Does this still work if you remove this config? Do we have to make this change somewhere else?

src/theme/Toggle/index.js
```diff
@@ -1,77 +0,0 @@
-/**
```
eak12913, line 1: Were these not being customized in spite of having been swizzled?

eak12913 (commented):
>Docusaurus now requires node v16.14.0; do we need to make infra changes for stage & prod?

I don't think we will need to make changes beyond what you did in CircleCI config. There's no `node` server running in prod. It's just generating a bunch of HTML files that are being uploaded into an S3 bucket.

docusaurus.config.js
oredavids: This worked hand-in-hand with the swizzled Toggle Component. Now that we are no longer swizzling it; the config is not required.

src/theme/Toggle/index.js
oredavids: The component was swizzled solely to customize the appearance via CSS. See from line 107 of the also deleted stylesheet [src/theme/Toggle/styles.module.css](https://github.com/gruntwork-io/docs/blob/f32a0783814c57feab50ea4746eab50a034ea0c2/src/theme/Toggle/styles.module.css#L107).

I showed @ebeneliason the new toggle in this version of Docusaurus and we agreed that there is no need to swizzle anymore as the new appearance is in line with what we implemented.

oredavids:
> I don't think we will need to make changes beyond what you did in CircleCI config. There's no node server running in prod. It's just generating a bunch of HTML files that are being uploaded into an S3 bucket.

Ah, yes! Thank you.

eak12913 (approved):
LGTM!

PR #639, opened 2023-02-15
[CORE-522] Cleanup deleted pages in S3 bucket

oredavids:
- Uses `aws s3 sync` option for deploys so that deleted files in source get removed from the s3 bucket
- Removes stubs and placeholder pages

[CORE-522]

[CORE-522]: https://gruntwork.atlassian.net/browse/CORE-522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


oredavids:
The config contained in the files where changes were made are safe to check in.

eak12913 (approved):
LGTM

eak12913 (approved):
LGTM - but the "stubs" in the modules weren't by accident. I think we left those in on purpose as we had the intention to put our generated docs there. Either way - makes sense to remove for now and then I can just fill that in when I ship my generated module docs.

PR #648, opened 2023-02-23
Update Production Framework guide to remove roadmap items

josh-padnick:
[A customer reported confusion](https://gruntwork-community.slack.com/archives/CHH9Y3Z62/p1677090694385829) on a roadmap commitment we made here. We shouldn't be making roadmap commitments here since it's too hard to keep track of them and they'll inevitably get out of date.

So I've updated the "How Gruntwork Helps" column to reflect our latest offerings and removed the "On the Roadmap" column.


eak12913 (approved):
LGTM! Thank you Josh!

PR #649, opened 2023-02-23
Update callout to point to new Gruntwork Pipelines example

zackproser:
Set-Up-an-Infrastructure-CI-CD-Pipeline-Gruntwork-Docs

These changes update the callout on our DIY Pipelines guide to reflect: 
1.The fact that the DIY Pipelines guide on the docs site is more of a high level design document 
2. We have an official Pipelines example in the service catalog now that is intended to be deployed

The callout also links to the official example in our service catalog.

Questions for reviewers:
1. Have I missed anything that would be good to update in light of the example? 
2. Are all the auto-generated hash updates in this PR expected? I was running the local dev server while editing `docs_sources`.

It should hopefully be a welcome change!

### DIFF
--- a/_docs-sources/guides/build-it-yourself/pipelines/index.md
+++ b/_docs-sources/guides/build-it-yourself/pipelines/index.md
@@ -7,24 +7,23 @@ import { CardList } from "/src/components/CardGroup"
 
 # Set Up an Infrastructure CI/CD Pipeline
 
-:::caution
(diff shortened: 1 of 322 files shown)


zackproser:
I think I see what's up - possible formatting issue causing unnecessary file changes to be picked up. I'll investigate and re-open a hopefully smaller PR.

PR #650, opened 2023-02-23
Update callout to reflect new Pipelines example

zackproser:
Set-Up-an-Infrastructure-CI-CD-Pipeline-Gruntwork-Docs

These changes update the callout on our DIY Pipelines guide to reflect: 
1.The fact that the DIY Pipelines guide on the docs site is more of a high level design document 
2. We have an official Pipelines example in the service catalog now that is intended to be deployed

The callout also links to the official example in our service catalog.

Questions for reviewers:
1. Have I missed anything that would be good to update in light of the example? 

It should hopefully be a welcome change!


robmorgan (approved):
LGTM

PR #652, opened 2023-02-23
[CORE-499][CORE-500] Added modules reference

eak12913:
This PR adds a new modules reference area to the docs site that exposes the READMEs and the inputs and outputs for each of the modules included in a standard Gruntwork subscription.

This PR also introduces comment parsing for our `variables.tf` files. This allows us to expose comments and examples that were previously only available to customer if they dug through the code. Now these comments and examples are available in a more first-class way to make them easier for customers to discover.


ebeneliason (approved):
Some notes based on my review. I don't think any of these are blockers, but some feel worth prioritizing.

* The "next" button on the intro page is a self-referencing link that goes nowhere.
* Release notes currently link to the general release notes for the repo, not those for the chosen module.
* Some modules have only required (or only optional) inputs, which feels a little odd. But maybe we can correct that later with a holistic approach to categorization including required optional as well as variable groupings.
* Fargate module is essentially a redirect.
* All of the cross-links to other modules go to GitHub — they really should go to the module page on the docs site.

eak12913:
I've noted all of the other issues for follow up! Merging now

PR #659, opened 2023-02-28
Clean up Modules intro section

eak12913:
A few small fixes:
- Cleaned up the text that explains modules vs services
- Removed the "next" link that went no where.

### DIFF
--- a/_docs-sources/reference/modules/intro.md
+++ b/_docs-sources/reference/modules/intro.md
@@ -3,8 +3,17 @@
 # Fortunately, this property only seems to affect # auto-generated sidebar items.
 # So we still _do_ display this page in the sidebar, but use a manual entry.
 sidebar_class_name: dont_show
+pagination_next: null
 ---
 
 # Gruntwork Module Catalog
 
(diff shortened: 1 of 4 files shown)


_docs-sources/reference/modules/intro.md
```diff
@@ -3,8 +3,17 @@
...
+
+**Services:** Reusable code that combines multiple modules to configure a service for a specific use case. Services are designed for specific use cases and meant to be deployed directly. For example, the eks-cluster service combines all the modules you need to run an EKS (Kubernetes) cluster in a typical production environment, including modules for the control plane, worker nodes, secrets management, log aggregation, alerting, and so on. The Gruntwork Service Catalog is a collection of battle-tested, commercially supported and maintained services that you can use to deploy production-grade infrastructure in minutes.
```
marinalimeira, line 19: Maybe link to the https://github.com/gruntwork-io/terraform-aws-service-catalog?

marinalimeira (approved):
LGTM

_docs-sources/reference/modules/intro.md
```diff
@@ -3,8 +3,17 @@
 # Fortunately, this property only seems to affect # auto-generated sidebar items.
 # So we still _do_ display this page in the sidebar, but use a manual entry.
 sidebar_class_name: dont_show
+pagination_next: null
```
eak12913, line 6: This is to fix a bug that was reported where the "next" link from this page just went back to itself. In this case - we actually don't want any "next" link showing up at all.

marinalimeira (approved):

PR #661, opened 2023-03-01
chore(CORE-521): Temporarily hide the cicd pipeline guide

MoonMoon1919:
### What it is

Temporarily removes the existing DIY guide for GW pipelines by removing references to the content in the sidebar.

I recognize that the content will still be available, if you want to seek out the links, but this seemed like the lower risk approach while we work on re-writing this guide. It also allows us to re-use parts of it without needing to comb through git history.

Happy to remove it completely, if that's the consensus of the team.


bethadele:
This works for me if Zack is a 👍🏻 . Thanks!

zackproser:
LGTM - I think this is a reasonable approach for now!

PR #668, opened 2023-03-03
CIS v1.5 Upgrade - text fixes

marinalimeira:
While recording the demo, I noticed somethings were incomplete:

1. Improved Steampipe configuration for multi-region access
1. Added the manual fixes that are not related to the "manual steps" page
1. Added more information regarding running patcher (that should be in each env folder), and about running the generated script.


brikis98 (approved):
These are nice improvements 👍

marinalimeira:
Thanks for the review!

PR #679, opened 2023-03-15
Remove placeholder pages from _docs-sources folder

oredavids:
This prevents a copies being generated in the `docs` directory when the local-copier plugin runs during `yarn start`.


eak12913 (approved):
LGTM

eak12913 (approved):
stamp

PR #687, opened 2023-03-21
Updated min. CIS repo version required for Patcher - PATCHER-116

techpink:
Changed min version in step 2 from `v0.40.0` to `v0.40.1`. Patcher v0.1.1 requires the version to be `>= v0.40.1`.


eak12913:
Hmm - for this PR, it looks like our `docs-sourcer` hook didn't run on it and didn't actually produce an output file into the `docs/` directory.  I think this is because the use case we covered in the `docs-sourcer` was a situation where the file contents were updated in the `docs/..../..../..../file_you_changed.md` and needed to be also synced back into `docs_sources/..../..../..../file_you_changed.md`

However - it doesn't look like we handle the other way around...as in the case of this PR. So what still needs to happen here is that someone needs to run the following yarn command locally in the `docs` project: `yarn regenerate:local`. Then you should have an updated file in the `docs/guides/stay-up-to-date/cis/cis-1.5.0/deployment-walkthrough/step-2-update-references-to-the-gruntwork-infrastructure-as-code-library.md` that needs to be checked in.

Please let me know if that makes sense - if not - happy to hop on and pair to demonstrate.

MoonMoon1919 (approved):

PR #791, opened 2023-05-17
Feat(CORE-942): Add staying up to date docs for IaC library

MoonMoon1919:


docs/iac/stay-up-to-date/versioning.md
```diff
@@ -0,0 +1,38 @@
+# Versioning
+
+Gruntwork versions the IaC library using [Semantic Versioning](https://semver.org/) (SemVer). Since much of the Gruntwork IaC Library is still pre-1.0.0, most of the Gruntwork IaC Library uses 0.MINOR.PATCH version numbers. With 0.MINOR.PATCH, the rules are a bit different, where we increment the:
+
+- MINOR version when you make backward incompatible API changes, and
+- PATCH version when you add backward compatible functionality or bug fixes.
+
+For modules that have submodules (e.g., terraform-aws-server/modules/single-server), not every release contains changes to every module. While using the latest available version is recommended, the version that most recently contains changes for a module can be found in each submodules reference in the [Library Reference](../reference/index.md).
+
+New module versions are released by creating a release on Github.  When a new version is created, refer to the release notes for a list of changes and migration guides (when necessary).
```
zackproser, line 10:
```suggestion
New module versions are released by creating a release on Github. When a new version is created, refer to the release notes for a list of changes and migration guides (when necessary).
```

zackproser:
Looking good locally! Couple of thoughts to consider or reject as you see fit: 

1. Are you thinking of intentionally keeping things very high level - or do you think we should explain that you also need to terraform and/or terragrunt apply successfully to make your update live?
2. Do you think we should get into, either on this doc or a separate future one, what is involved with Terraform / Terragrunt binary version updates? I'm coming at this as if I were a semi-new user of Terraform / IaC - would I want a heads up about the fact that certain state buckets will have been applied with a certain version of Terraform and what that means for upgrading?

MoonMoon1919:
> 1. Are you thinking of intentionally keeping things very high level - or do you think we should explain that you also need to terraform and/or terragrunt apply successfully to make your update live?

I intentionally left it high level but this seems like a good thing to add. It _is_ required, after all.

MoonMoon1919:
> 2\. Do you think we should get into, either on this doc or a separate future one, what is involved with Terraform / Terragrunt binary version updates? I'm coming at this as if I were a semi-new user of Terraform / IaC - would I want a heads up about the fact that certain state buckets will have been applied with a certain version of Terraform and what that means for upgrading?

I like the concept, since we usually pin a specific terraform version in a module. I'm wondering if would be better in a blog post/learning center post. It's related to the IaC library since you _have_ to use one of those two tools, but upgrading the tools themselves is not necessarily required information just to use the library the majority of the time.

eak12913 (approved):
LGTM

PR #835, opened 2023-05-30
Feat(CORE-976): add terragrunt and _envcommon update and version guides

MoonMoon1919:


_docs-sources/iac/stay-up-to-date/updating.md
```diff
@@ -29,6 +35,97 @@ module "my_instance" {
...
+```
+
+After making the change, run `terragrunt plan`, inspect the output to ensure it looks as you expect, then run `terragrunt apply`.
+
+</TabItem>
+<TabItem value="Terragrunt with _envcommon" label="_envcommon (Terragrunt)">
+
+Below is an example using the `_envcommon` pattern to reference version `0.15.3` of the `single-server` submodule from the `terraform-aws-server` module.
+
+To update to version `0.15.4`, you update the value to the right of `ref=` in the source attribute. Since the version number denotes that this update is backwards compatible, it should not require any other changes. When following the `_envcommon` pattern, there are two places that reference the git tag created by the release — the `.hcl` file with the reference to the module in the `_envcommon` directory, and the environment and region specific reference to that file.
```
zackproser, line 81:
```suggestion
To update to version `0.15.4`, you update the value to the right of `ref=` in the source attribute. Since the version number denotes that this update is backwards compatible, it should not require any other changes. When following the `_envcommon` pattern, there are two places that reference the git tag created by the release — the `.hcl` file with the reference to the module in the `_envcommon` directory, and the environment and region specific reference to this _envcommon file.
```

zackproser: Not sure if this actually makes the sentence more clear or not, but thought I'd try it out. Feel free to reject.

zackproser (approved):
LGTM!

_docs-sources/iac/stay-up-to-date/updating.md
MoonMoon1919: Good call - I updated to `specific references to the _envcommon file`

zackproser: Much more clear - thanks!

zackproser (approved):
LGTM!

zackproser (approved):
LGTM

PR #838, opened 2023-05-30
Feat(CORE-941): Add using the library page

MoonMoon1919:
## What it is

Adds `Using the library` page to the IAC usage docs


_docs-sources/iac/usage/using-the-library.md
```diff
@@ -0,0 +1,396 @@
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Using the Library
+
+The Terraform modules in the Gruntwork Infrastructure as Code Library allows you to configure the `provider` and `backend` settings to suit their needs. This makes it possible to use Gruntwork modules alongside any existing modules you may have already developed, with minimal duplication of configuration.
```
zackproser, line 6:
```suggestion
The Terraform modules in the Gruntwork Infrastructure as Code Library allow you to configure the `provider` and `backend` settings to suit your needs. This makes it possible to use Gruntwork modules alongside any existing modules you may have already developed, with minimal duplication of configuration.
```

_docs-sources/iac/usage/using-the-library.md
```diff
@@ -0,0 +1,396 @@
...
+- [Terraform](https://www.terraform.io) installed on your local machine
+- (Optional — only required for testing) [Go](https://go.dev) installed on your local machine
+
+## Create a module
+
+In this section you’ll create a module that provisions an AWS Lambda Function using the [`terraform-aws-lambda`](../../reference/modules/terraform-aws-lambda/lambda/) Gruntwork module. This module automatically creates the AWS IAM role and Cloudwatch Log Group for the Lambda function. For more information about the configuration of the module, refer to the [Library Reference](../../reference/modules/terraform-aws-lambda/lambda/#reference).
+
+### Create the basic file structure
+
+First, create the basic file structure required that will contain the module reference. In this guide, you will create a module named `serverless-api` which contains the reference to the `terraform-aws-lambda` module. This approach is taken to allow you to define a module once, then use it in many environments and regions, while keeping the amount of copy-and-paste work to a minimum.
```
zackproser, line 24:
```suggestion
First, create the basic file structure that will contain the module reference. In this guide, you will create a module named `serverless-api` which contains the reference to the `terraform-aws-lambda` module. This approach is taken to allow you to define a module once, then use it in many environments and regions, while keeping the amount of copy-and-paste work to a minimum.
```

zackproser: NIT: passive voice in "This approach is taken..."

_docs-sources/iac/usage/using-the-library.md
```diff
@@ -0,0 +1,396 @@
...
+</Tabs>
+
+### Create the reference to the Gruntwork module
+
+Next, create the reference to the `terraform-aws-lambda` module...
+
+<Tabs groupId="tool-choice">
+<TabItem value="Terraform" label="Terraform" default>
+
+Define a module block that uses the git url `terraform-aws-lambda` module for the `source` attribute.
```
zackproser, line 72:
```suggestion
Define a module block that uses the git url of the `terraform-aws-lambda` module for the `source` attribute.
```

_docs-sources/iac/usage/using-the-library.md
```diff
@@ -0,0 +1,396 @@
...
+  type        = string
+  description = "The name of the handler function that will be called as the entrypoint of the lambda"
+}
+
+```
+
+</TabItem>
+<TabItem value="Terragrunt" label="Terragrunt" default>
+
+Define the module in `_envcommon` that references the `terraform-aws-lambda`. For `terragrunt`, the source is defined in the `source` attribute of the `terraform` block.
```
zackproser, line 118:
```suggestion
Define the module in `_envcommon` that references the `terraform-aws-lambda` module. For `terragrunt`, the source is defined in the `source` attribute of the `terraform` block.
```

zackproser: Feels like an overuse of the word module, but without it, it also sounds strange to me.

_docs-sources/iac/usage/using-the-library.md
```diff
@@ -0,0 +1,396 @@
...
+
+One of the benefits of referencing modules this way is that ability to set defaults for your organization. As an example — the `terraform-aws-lambda` exposes many variables but in the module block below, we are hard coding the value `run_in_vpc` to be `false`. This will ensure that anyone consuming this module will only create AWS Lambda functions that are not in a VPC. For a full list of configuration options for this module, refer to the [Library Reference](../../reference/modules/terraform-aws-lambda/lambda/#reference).
```
zackproser, line 120:
```suggestion
One of the benefits of referencing modules this way is the ability to set defaults for your organization. As an example — the `terraform-aws-lambda` exposes many variables but in the module block below, we are hard coding the value `run_in_vpc` to be `false`. This will ensure that anyone consuming this module will only create AWS Lambda functions that are not in a VPC. For a full list of configuration options for this module, refer to the [Library Reference](../../reference/modules/terraform-aws-lambda/lambda/#reference).
```

_docs-sources/iac/usage/using-the-library.md
```diff
@@ -0,0 +1,396 @@
...
+def lambda_handler(event, context):
+    return "Hello from Gruntwork!"
+```
+
+</TabItem>
+<TabItem value="Terragrunt" label="Terragrunt" default>
+
+Create a module block that uses the path to the local module as the source attribute, supplying values for the required attributes of the module.
+
+One of the benefits of this approach is that you can increment the version of a module for a specific environment and region in a granular fashion. For example, when we shipped a version v0.22.0 of the `terraform-aws-lambda` module, you could update just the `example` environment in the us-west-2 AWS region to ensure the upgrade goes as expected, then roll out to other environments or regions.
```
zackproser, line 173:
```suggestion
One of the benefits of this approach is that you can increment the version of a module for a specific environment and region in a granular fashion. For example, if we were to ship a version v0.22.0 of the `terraform-aws-lambda` module, you could update just the `example` environment in the us-west-2 AWS region to ensure the upgrade goes as expected, then roll out to other environments or regions.
```

_docs-sources/iac/usage/using-the-library.md
MoonMoon1919: Good catch - this is what i meant it to say :lolsob:

_docs-sources/iac/usage/using-the-library.md
```diff
@@ -0,0 +1,396 @@
...
+Finally, run the test you wrote. From the `test` directory, run the following command:
+```bash
+go test -v
+```
+
+You should expect to see `--- PASS: TestLambdaCreated` in the final log lines of the output from the test.
+
+## What’s next
+
+Now that you’ve used a Gruntwork module to provision resources, consider how you would need to expand this usage make the Lambda function available via a URL using an [AWS API Gateway HTTP API](../../reference/modules/terraform-aws-lambda/api-gateway-proxy/). We refer to combining multiple modules into a single deliverable module as a [service](../overview/services.md).
```
zackproser, line 391:
```suggestion
Now that you’ve used a Gruntwork module to provision resources, consider how you would need to expand this usage to make the Lambda function available via a URL using an [AWS API Gateway HTTP API](../../reference/modules/terraform-aws-lambda/api-gateway-proxy/). We refer to combining multiple modules into a single deliverable module as a [service](../overview/services.md).
```

_docs-sources/iac/usage/using-the-library.md
```diff
@@ -0,0 +1,396 @@
...
+
+<!-- Leaving this commented out since these two docs pages are coming soon -->
```
zackproser, line 393:
```suggestion
```

zackproser:
Left some nits but otherwise this is looking great! Nicely done.

_docs-sources/iac/usage/using-the-library.md
MoonMoon1919: Removed in https://github.com/gruntwork-io/docs/pull/838#:~:text=Verified-,5d6ac59,-Merge%20state

_docs-sources/iac/usage/using-the-library.md
MoonMoon1919: I did some re-ordering of the wording here, moving the common paragraph up above into the section header. LMK what you think.

_docs-sources/iac/usage/using-the-library.md
```diff
@@ -0,0 +1,391 @@
...
+	// Run this test in parallel with all the others
+	t.Parallel()
+
+	// Unique ID to namespace resources
+	uniqueId := random.UniqueId()
+	// Generate a unique name for each Lambda so any tests running in parallel don’t clash
+	lambdaName := fmt.Sprintf("test-lambda-%s", uniqueId)
+
+	// Get the cwd so we can point to the lambda handler
+	path, _ := os.Getwd()
```
zackproser, line 341: NIT: handle the error here

zackproser (approved):
Max, this is terrific.

zackproser (approved):
LGTM!

ebeneliason:
* "wrapper module" link in file structure section doesn't work
* is it sufficient to suggest `plan`/`apply` here, or should we also call out how one would leverage this in a GitOpsPR workflow?
* "We recommend putting all tests in the test directory in your repository." Is that strong enough? I know we cover the structure of modules and services in their respective intro sections, but I wonder whether we should be doing more to imply that this is our canonical best-practices structure. Should it get a section unto itself in the sidebar? Even if not, should we say something like: "We recommend putting all tests in the `test` directory in your repository in accordance with our [recommended module structure](…)"?
* Testing, here, is just one section of this guide. I wonder, even if that remains true, should we not have a top level section under "Working with the library" on "Testing your modules"? It feels like a topic unto itself. Having that would also allow us to separate the "here's how you should approach testing your modules" stuff from the "now copy and run this bit of code" imperative approach.

MoonMoon1919:
> * "wrapper module" link in file structure section doesn't work

Will update, thanks.

> * is it sufficient to suggest `plan`/`apply` here, or should we also call out how one would leverage this in a GitOps PR workflow?

I'm inclined to add a few sentences or a paragraph about it in the "What's next". This is a guide that's meant to be followed top to bottom, so it might be awkward to try to slot in "hey you could put this in CI" in the middle of it. Thoughts?

> * "We recommend putting all tests in the test directory in your repository." Is that strong enough? I know we cover the structure of modules and services in their respective intro sections, but I wonder whether we should be doing more to imply that this is our canonical best-practices structure. Should it get a section unto itself in the sidebar? Even if not, should we say something like: "We recommend putting all tests in the `test` directory in your repository in accordance with our [recommended module structure](%E2%80%A6)"?

I can re-word this section.

> * Testing, here, is just one section of this guide. I wonder, even if that remains true, should we not have a top level section under "Working with the library" on "Testing your modules"? It feels like a topic unto itself. Having that would also allow us to separate the "here's how you should approach testing your modules" stuff from the "now copy and run this bit of code" imperative approach.

I like the idea of including testing as a separate section, but not in the scope of this work.

zackproser (approved):
LGTM!

PR #840, opened 2023-05-31
Fix issues associated with running locally

josh-padnick:
I wanted to run Docusaurus locally so I could more easily propose updates to the content, but I kept getting dozens of instances of the following error:

```
Error: error:0308010C:digital envelope routines::unsupported
    at String.replace (<anonymous>). This is apparently an issue due to using vulnerable dependencies that rely on outdated TLS standards.
```

Based on [this Stack Overflow response](https://stackoverflow.com/a/73027407), I ran `yarn-audit-fix`, the errors went away, and our dependencies are patched to later versions.

I haven't done any additional testing beyond that, so I'm not sure if there are undesirable side effects from this change.


josh-padnick:
FYI, the `.gitignore` config I'm using doesn't seem to be working.

.gitignore
```diff
@@ -84,3 +84,4 @@ build/
 
 # VS Code files
 .vscode
+docs/guides/stay-up-to-date/index.md
```
eak12913, line 87: We actually don't want to do this as that will make it much harder to change anything on this page if we ever want to do that. 

Instead we need to either architect the `docs-sourcer` to allow this kind of functionality (ie - mixing auto-generated and manually written content on a page) or we need to update our IA such that this content gets separated.

package.json
```diff
@@ -50,7 +50,8 @@
     "ts-node": "^10.7.0",
     "tsconfig-paths": "^3.14.1",
     "typescript": "^4.3.5",
-    "yargs": "^17.4.0"
+    "yargs": "^17.4.0",
+    "yarn-audit-fix": "^9.3.10"
```
eak12913, line 54: Once you've run `yarn-audit-fix`, it has changed your `lock` file. Is there any point of keeping this dependency around?

josh-padnick: Fair enough! Removed in 1e9c4202ac2c8ef65f2ca4482db6d742f73efe5b

.gitignore
josh-padnick: Ah, good point. Strangely, this `.gitignore` didn't work anyway. I think having a fully auto-generated page is the way to go.

Separately, it looks like https://docs.gruntwork.io/guides/stay-up-to-date/#gruntwork-releases is now hidden from the IA. Perhaps we should add a new page under Library called "Release Notes" that includes the releases section in https://docs.gruntwork.io/guides/stay-up-to-date/#gruntwork-releases?

josh-padnick:
@eak12913 This is now just a lock file change. Could you either approve and merge this, or close it out so we can remove this as a work-in-progress PR?

marinalimeira (approved):

PR #841, opened 2023-05-31
Rename Gruntwork IaC Library to Gruntwork Library

josh-padnick:
This is a quick first pass at renaming Library to start the conversation. 

⚠️ If we decide to move forward, we'll need to make a few other changes:

- Renaming `/iac/` to `/library/`
- Doing a global search and replace

## Why rename Gruntwork IaC Library to Gruntwork Library?

Now that we've narrowed our focus to DevOps Foundations, I find it helpful to think in terms of our "component" products, mainly:

- Reference Architecture
- Pipelines
- Library

In this context, each component is a collection of functionality designed to do one thing well. In the case of Library, it's a way for us to capture best practices and share them with our customers.

These best practices include infrastructure as code, but they'll include other things in the future, too, such as boilerplate templates, or policies, or Patches. It's a bit of a stretch to consider all of these things as "infrastructure as code" so it's cleaner to refer to this part of our product as just "Library" and to have that mean "All the best practices in DevOps that Gruntwork recommends," whether those practices are Terraform code or something else.

As a side benefit, "Library" is easier to say than "The IaC Library," and fits better in the sidebar and nav.

image
image

image
image

@ebeneliason This is a public PR so let's discuss this via internal Slack, but I started making these changes and realized they needed to be packaged as a separate PR so we could make these changes as a single transaction.


ebeneliason:
> If you're committed to this approach for the long term then is your preference to just make the textual changes immediately, and then schedule the more onerous work of renaming paths and adding redirects?

Sanity check: didn't we just add these new paths last sprint? If we change them promptly, do we need to go through the whole redirect process? There's been barely any time at all for these new paths to build up SEO. If we do have to carve out a separate story for managing paths, I'm still inclined to clean up the language sooner anyway.

MoonMoon1919:
> > If you're committed to this approach for the long term then is your preference to just make the textual changes immediately, and then schedule the more onerous work of renaming paths and adding redirects?
> 
> Sanity check: didn't we just add these new paths last sprint? If we change them promptly, do we need to go through the whole redirect process? There's been barely any time at all for these new paths to build up SEO. If we do have to carve out a separate story for managing paths, I'm still inclined to clean up the language sooner anyway.

Yes - these are all new paths. I think given the amount of time that they've been up (less than 2 weeks) we'll be okay without the redirects.

josh-padnick:
Great! @ebeneliason Could you figure out how to get this story across the finish line?

josh-padnick:
@ebeneliason Could you work with the team to prioritize this as needed?

ebeneliason:
I haven't forgotten this, but it has been on the back burner. I was hoping to address it via the new Docs Jams RFC proposal, but that RFC in itself has been put on hold while we finish cycle planning. Regardless of whether we move forward with that proposal, we need a good answer to ensuring PRs like this don't rot.

josh-padnick:
Totally understood; we have a full plate right now. Since this is a flex week, does it make sense to select an engineer and ask them to run with it?

ellisonc (approved):

ellisonc (dismissed):

PR #855, opened 2023-06-05
Core-977 - inject GitHub username into lambda handler and call Mixpanel API directly

zackproser:
Relates to: https://github.com/gruntwork-io/dogfood-infrastructure-modules/pull/150

These changes examine the URL's query string params for a github_username and inject it, along with our tutorial service's endpoint, into the user's demo lambda handler code. 

The injections are implemented by a new custom client-side script included in this PR. The client side script is designed to *only* run on our "Deploying your first module" tutorial page, so it should theoretically have minimal impact on the operation of the rest of the docs site. 

The injected endpoint is our GW tutorial service's API gateway deployment URL. The customer's demo lambda's Python handler code has also been updated in this PR to call this specific endpoint and return its response, thus completing the round-trip and book-keeping for tracking activations / users who reached our tutorial goal state.

Ideally, I'd like to figure out how to properly define our GW tutorial service's endpoint solely in `docusaurus.config.js` and load it from within the custom client script that handles the value injection on our "Deploying your first module" , but currently when I attempt to follow docusaurus's docs on loading docusaurus config values I get an error about misusing React hooks. I'd like to manage this config centrally so that future PRs to update the endpoint could be one line change.


_docs-sources/iac/getting-started/deploying-a-module.md
```diff
@@ -144,8 +144,26 @@ Next, we’ll write a simple Python function that returns a string that will be
 Copy the following to `terraform-aws-gw-lambda-tutorial/main.py`.
 
 ```py title="terraform-aws-gw-lambda-tutorial/main.py"
+import json
+from urllib.request import urlopen, Request
```
MoonMoon1919, line 148: We might be able to simplify this using `requests`, which is an OSS package with a nicer interface than urllib. At one point was available in the python lambda runtime like so - `from botocore.vendored import requests`.

At one point AWS was considering removing it from the runtime, but it seems like they cancelled that work. Worth a try to reduce the number of lines of code?

MoonMoon1919 (commented):
Small, non-blocking nit, but otherwise LGTM

_docs-sources/iac/getting-started/deploying-a-module.md
zackproser: Very cool. Will give it a shot - I was indeed trying to keep this handler as minimal as possible - and hopefully the next pass I'll put up shortly will do just that!

zackproser: Alrighty - with the latest changes I can confirm everything works as intended end to end, and I think the resulting lambda handler code, while a bit longer, is not too overwhelming.

MoonMoon1919: excellent, sorry for the rabbit hole!

_docs-sources/iac/getting-started/deploying-a-module.md
```diff
@@ -144,8 +144,57 @@ Next, we’ll write a simple Python function that returns a string that will be
...
 ```py title="terraform-aws-gw-lambda-tutorial/main.py"
+import uuid
+import base64
+import json
+from urllib.request import urlopen, Request
+
+
 def lambda_handler(event, context):
-    return "Hello from Gruntwork!"
+    url = "https://api.mixpanel.com/track"
```
MoonMoon1919, line 154: 👨🏻‍🍳 🤌 excellent

scripts/module-tutorial-client.js
```diff
@@ -0,0 +1,41 @@
...
+// 
+// Separately, regardless of whether or not the user's GitHub username is present in the query params, we must base64 encode our 
+const injectGitHubUsernameIfPresent = () => {
+  const urlParams = new URLSearchParams(window.location.search);
+  let githubUsername = urlParams.get('github_username');
+  if (githubUsername == null) {
+    githubUsername = 'unknown'
+  }
+  // Rewrite the %unknown% placeholder in the main.py configuration on the tutorial page
+  document.body.innerHTML = document.body.innerHTML.replace('%unknown%', githubUsername);
```
MoonMoon1919, line 16: slick!

docusaurus.config.js
```diff
@@ -65,14 +66,22 @@ const config = {
...
       },
     ],
   ],
 
+  // clientModules are custom modules (anything that Webpack can import), so css, js, etc
+  // that get included on every page and can be used to run arbitrary client-side Javascript, for example: 
+  clientModules: [path.resolve('./scripts/module-tutorial-client.js')],
+
+  customFields: {
+    tutorialServiceEndpoint: "https://h1t1m9ck07.execute-api.us-east-1.amazonaws.com/tutorial-service/post"
```
MoonMoon1919, line 82: Is this still being used somewhere?

MoonMoon1919 (commented):
Looking great. Did you check in `package-lock.json` by accident, or is that a bi-product of something else?

docusaurus.config.js
zackproser: Ah great catch - thank you! Will remove.

zackproser: Removed

zackproser:
> Looking great. Did you check in `package-lock.json` by accident, or is that a bi-product of something else?

Ah yeah - that was actually intended: in the docusaurus config file, to load the new client module script, I needed to call `path` - but seems I may not have done that correctly yet - I was originally getting build errors about Webpack > 5 having removed polyfills for certain node.js APIs such as path - so I thought the fix was to explicitly npm install it. 

Not sure yet if current build errors are still related to that or not. Yesterday and the day before, Netlify logs were showing me an auth error where it looked like Netlify was unable to fetch the repo itself during the build :thinking:

package.json
```diff
@@ -32,6 +32,7 @@
     "config": "^3.3.6",
     "env-cmd": "^10.1.0",
     "file-loader": "^6.2.0",
+    "path": "^0.12.7",
```
MoonMoon1919, line 35: Do you also need to `yarn add buffer` for the module-tutorial-client package to work?

MoonMoon1919 (approved):

zackproser:
Thanks for review!

PR #874, opened 2023-06-13
Add script to create releases PR

MoonMoon1919:
Adds script to automatically generate releases and create a PR

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,12 +21,28 @@ references:
   # Build steps
   ######################################################################################################################
 
+  install_gh: &install_gh
+    run:
+      name: install gh cli
+      command: |
+        type -p curl >/dev/null || (sudo apt update && sudo apt install curl -y)
+        curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
+        && sudo chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
+        && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
+        && sudo apt update \
+        && sudo apt install gh -y
+
   install_bash_commons: &install_bash_commons
     run:
       name: install bash commons
       command: |
         curl -Ls https://raw.githubusercontent.com/gruntwork-io/gruntwork-installer/master/bootstrap-gruntwork-installer.sh | bash /dev/stdin --version v0.0.21
         gruntwork-install --module-name bash-commons --repo https://github.com/gruntwork-io/bash-commons --tag "v0.1.2"
+
+  install_deps: &install_deps
+    run:
+      name: install deps
+      command: yarn install
 
   build_docs: &build_docs
     run:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -64,6 +80,14 @@ references:
         export ALGOLIA_LIBRARY_INDEX_NAME="prod_docs_sourcer-library-reference"
         ./scripts/push-to-s3-prod.sh
 
+  create_releases_pr: &create_releases_pr
+    run:
+      name: create releases pr
+      command: ./scripts/create_releases_pr.sh
+      environment:
+        CONFIGURE_GIT: 0 # Tell the script that we're in CI and need to configure git for gruntwork-ci user
```
eak12913, line 88: Why `0`? Can we not pass a boolean?

MoonMoon1919: we can pass a string that says "true" or "false" but we'll have to treat it as a string. I'm fine either way.

.circleci/config.yml
```diff
@@ -197,3 +237,13 @@ workflows:
               only: /^v.*/
             branches:
               ignore: /.*/
+  create_releases_pull_request:
+    when:
+      and:
+        - equal: [scheduled_pipeline, << pipeline.trigger_source >>]
```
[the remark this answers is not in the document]
eak12913: I intuitively understand that this is the "cron" equivalent for CircleCI which makes this fire off monthly - but I'm missing a bunch of details that would give a better understanding of what specifically is happening

scripts/create_releases_pr.sh
```diff
@@ -0,0 +1,100 @@
...
+export RELEASES_PLUGIN_BEFORE_DATE=
+export BRANCH_NAME=
+export CONFIGURE_GIT=${CONFIGURE_GIT:=1}
+
+function set_release_date() {
+  unameOut="$(uname -s)"
+
+  # Handling running this on linux or mac
+  if [ ${unameOut} = "Linux" ]; then
+    RELEASES_PLUGIN_BEFORE_DATE=$(date -d "$(date +"%Y-%m-01") - 1 day" +"%F")
```
eak12913, line 17: Is there any issue with having nested double quotes like this?

scripts/create_releases_pr.sh
```diff
@@ -0,0 +1,100 @@
...
+    exit 1
+  fi
+}
+
+function generate_release_data() {
+  yarn regenerate --plugins releases
+}
+
+function create_branch() {
+  BRANCH=$1
```
eak12913, line 53: `local -r` ? Also - why caps?

scripts/create_releases_pr.sh
```diff
@@ -0,0 +1,100 @@
...
+
+  git checkout -b ${BRANCH}
+}
+
+function add_and_commit_changes() {
+  RELEASE_DATE=$1
```
eak12913, line 59: `local -r`

scripts/create_releases_pr.sh
MoonMoon1919: It works the same with single quotes in there (i have to use gdate on my mac to get the gnu date syntax)

image


image

scripts/create_releases_pr.sh
MoonMoon1919: caps is the syntax i've always used for args in bash functions /shrug

scripts/create_releases_pr.sh
```diff
@@ -0,0 +1,100 @@
...
+
+function configure_git {
+  local -r git_username="$1"
+  local -r git_email="$2"
+  git config --global user.name "$git_username"
+  git config --global user.email "$git_email"
+  git config --global pull.rebase false
+}
+
+function set_branch_name() {
```
eak12913, line 34: consider whether you want to change to:
- Branch name isn't a global env var
- This function is `get_branch_name` which returns the formatted branch name
- You pass that as a param into the functions that need it.

It's not a big deal - but i found it weird that there was a global looking variable at the top that was not initialized.

Since this isn't shared with anything..nor is it any kind of reusable component, I don't think it _really_ matters..so this whole thing is just a nit.

.circleci/config.yml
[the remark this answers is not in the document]
MoonMoon1919: Added.

scripts/create_releases_pr.sh
MoonMoon1919: Updated.

oredavids (approved):
LGTM.

PR #897, opened 2023-06-16
Add self-hosting page

josh-padnick:
Add a page mentioning repo-copier as a way to self-host the library


docs/library/usage/self-hosting.md
```diff
@@ -0,0 +1,39 @@
...
+
+# Self-hosting the library
+
+:::caution
+
+Self-hosting Gruntwork Library is only available with a Gruntwork Enterprise subscription.
+
+:::
+
+We recommend accessing Gruntwork Library directly through our private repos hosted at https://github.com/gruntwork-io. However if your company requires that all Terraform modules are hosted internally, we offer a tool called [repo-copier](https://github.com/gruntwork-io/repo-copier) to support this use case.
```
oredavids, line 13:
```suggestion
Self-hosting the Gruntwork Library is only available with a Gruntwork Enterprise subscription.

:::

We recommend accessing the Gruntwork Library directly through our private repos hosted at https://github.com/gruntwork-io. However if your company requires that all Terraform modules are hosted internally, we offer a tool called [repo-copier](https://github.com/gruntwork-io/repo-copier) to support this use case.
```

oredavids (commented):
Content looks good. Left one minor suggestion.

The tests don't pass currently due to the `docs/guides/stay-up-to-date/index.md` changes. Please revert that. We unfortunately haven't worked on preventing that from happening when working locally.

oredavids:
Did not catch this earlier but the self-hosting markdown file needs to be moved into the `_docs-sources` directory and then run the `yarn regenerate:local` command so it is copied over to `docs` with its metadata.

docs/library/usage/self-hosting.md
josh-padnick: I actually prefer it without the "the" because I think of the product name as "Gruntwork Library" just like a product name lik Heroku, Render.com or Supabase.

In that way of thinking you wouldn't refer to "self-hosting the Heroku." I think "the Gruntwork Library" only sounds ok because you're describing a library whose name is "Gruntwork."

oredavids: Sounds good.

josh-padnick:
@oredavids I'm sorry, but I'm afraid I'm confused here. I keep trying variations of moving the file and running `yarn regenerate:local`. But It doesn't work.

oredavids:
@josh-padnick It appears that your last commit fixed the issues so i did not have to do anything.

josh-padnick:
Ah nice, thank you! Thought I'd already tried that, but guess not. Merging now!

PR #898, opened 2023-06-19
Update patcher changelogs as of 19. June

marinalimeira:
Data generated by running `DOCS_SITE_BASE_DIR="../docs/" yarn dev -p patcher-changelogs`.

⚠️ IMPORTANT: this PR removes the `logs` entry from `.gitignore`. [The `node` template from gitignore.io does create this file](https://github.com/toptal/gitignore/blob/36f6ed1fc974d610bd9fcd1e6fa656bf0ba1f10e/templates/Node.gitignore#L4). I will leave like this temporarily, if we need to re-introduce _some_ pattern to `logs`, we can add a better filter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ terraform.tfvars
 *.tfstate*
 
 # Logs
-logs
 *.log
 npm-debug.log*
 yarn-debug.log*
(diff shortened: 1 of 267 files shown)


oredavids (approved):
LGTM after quick sanity check on a small sample size.

marinalimeira:
Thanks a lot for the review!

PR #903, opened 2023-06-20
Add missing init step in using-the-library

MoonMoon1919:
## What it is

Adds explicit init step in using the library guide


oredavids (approved):
LGTM.

PR #1024, opened 2023-08-03
Updated docs to Patcher v0.4.2

techpink:
- Added a note that `patcher update` requires Docker by default.
- Updated the `patcher update --non-interactive` examples to include the `--skip-container-runtime` flag
- Updated the description of `patcher report` to include the non-interactive mode

This PR deliberately doesn't include information on how to use Patcher without Docker. This can't be documented until we've resolved how we are going to make terrapatch available.


marinalimeira (approved):
LGTM! 🚀

PR #1032, opened 2023-08-10
Add Patcher to the home page

ebeneliason:
This adds Patcher under the list of products on the home page, and removes Reference Architecture. A future update will shuffle some of the more prominent cards at the top as well.


marinalimeira (approved):
🚀

PR #1079, opened 2023-08-29
Adding envcommon docs

MoonMoon1919:
## What it is

Adds docs to explain _what_ the envcommon pattern is and _how_ to use it as both a module developer and consumer.

### DIFF
--- a/_docs-sources/learning-center/envcommon/defining.md
+++ b/_docs-sources/learning-center/envcommon/defining.md
@@ -44,9 +44,8 @@ inputs = {
     num_nat_gateways                          = 1
 }
 ```
+Our example is relatively straight forward, our module only has three required variables, but there are actually 79 more optional variables. As you expand your usage to supply values to more optional values, the amount of code you don't have to write would increase significantly.
 
 ## Wrapping up
-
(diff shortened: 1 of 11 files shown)


MoonMoon1919:
> * A learning center will feel underwhelming until we have some critical mass of content. I'm not sure how many topics we'd need to cover — maybe not that many, but definitely more than one.

Agreed!

> * We've already defined that Terraform Foundations section, but haven't yet filled it with content. This seems topically perfectly aligned.

Ah, I must have missed that. Will update, thanks!

> * We may introduce some form of curriculum, akin to the learning center concept, which focuses predominantly on the skills we _outsource_ to other learning resources (AWS, KodeKloud, etc.). In that model, we _may_ want to distinguish that content from the content that's specific to Gruntwork. (This falls right on the edge, since it's a TG topic, which we control, but is open source and kept at arm's length from our main docs. That said, envcommon isn't strictly TG-specific, but a Gruntwork-specific recommended usage of TG.)

Makes sense!

MoonMoon1919:
Thanks, @ebeneliason. Because our foundations branch is so far behind master, it wound up just being easier to create a new PR based off the devops-foundations branch. Please put all future comments in https://github.com/gruntwork-io/docs/pull/1082

PR #1101, opened 2023-09-08
Fix the path to RDS module in doc

hongil0316:
This issue has been raised by a customer here: https://gruntwork.zendesk.com/agent/tickets/110412.


MoonMoon1919:
@hongil0316 - Our docs are automatically generated, so we'll need to address this issue in that project.

PR #1114, opened 2023-09-14
Feat(CORE-1238): Add single account tutorial pipelines docs

MoonMoon1919:

PR #1116, opened 2023-09-18
Feat(CORE-1273): Add pipelines security docs

MoonMoon1919:

PR #1119, opened 2023-09-19
Add production use section

josh-padnick:
In https://github.com/gruntwork-io/docs/pull/1114, we tell the user how to get Pipelines up and running, but issue various warnings that for a production use case certain settings need to be tweaked.

This PR attempts to fill in those gaps. I don't have the knowledge needed to complete this, but have written out a set of stubs that hopefully make it easy to fill in the right information.

Please feel free to any other missing information necessary to take Pipelines from Hello World to ready for prod.


_docs-sources/pipelines/production-use/index.md
```diff
@@ -0,0 +1,27 @@
+# Production Use
+
+In the [Hello World Tutorial](../hello-world), we got Pipelines up and running for the first time. In this tutorial, we make it production-ready.
+
+## Branch protection rules
```
MoonMoon1919, line 5: This is duplicate work, I've already scaffolded this out here https://github.com/gruntwork-io/docs/pull/1116

josh-padnick: Actually, I think they are related but independent. The key business need here is to give customers step-by-step instructions for getting Pipelines ready for prod, including specific, opinionated branch protection rule recommendations (or calling out the options and tradeoffs among them).

By contrast, #1116 seems focused on explaining the mechanisms we use to achieve a strong security posture. 

To make the distinction more concrete, I could imagine the DevOps engineer tasked with setting up Pipelines referring to this "Production use" guide, while the Security engineer tasked with vetting Pipelines as acceptable to install in the first place, would review #1116.

MoonMoon1919: I recently updated #1116 to be more limited in scope after seeing a different ticket in the queue for documenting the token creation bits to reduce the potential for duplicate content, so I'm fine with distinguishing the two.

The only difference between the tutorial and a prod setup for pipelines is creating more tokens that have limited scope. I suspect this doc will be quite short.

MoonMoon1919: Added in #1126

josh-padnick: Roger, and thanks!

josh-padnick:
#1126 added info on machine users, and another PR added info on configuring branch protection, so the needs in this PR have now been addressed!

PR #1120, opened 2023-09-19
Add stubs for adding multiple AWS accounts.

josh-padnick:
Many users will have the need to set up Pipelines for more than one `infrastructure-live` repo, so we should address explicitly how users can do this.

Regarding priority, I'm not sure where this fits, but I would expects users to ask about it in the next few days, possibly as soon as Wednesday.


_docs-sources/pipelines/maintain/adding-aws-accounts.md
```diff
@@ -0,0 +1,16 @@
+# Adding AWS accounts
+
+If your organization has a need for more than one `infrastructure-live` repository, there are two primary approaches to adding Pipelines support to additional repositories:
+
+1. Create a new `infrastructure-pipelines` repository (though named something different) for the new `infrastructure-live` repository
+1. Use a single `infrastructure-pipelines` repository for multiple `infrastructure-live` repositories.
```
MoonMoon1919, line 6: This is not currently supported. We have a story this sprint to determine how to best accomplish this.

josh-padnick: Ah, that's going to be pretty timely, actually. I suppose in the meantime, customers can always use one Pipelines per infra-live repo, though.

josh-padnick:
It turns out that the ability to support multiple infra-live repos for one pipelines repo is part of [CORE-1274](https://gruntwork.atlassian.net/jira/software/c/projects/CORE/boards/89?selectedIssue=CORE-1274), whose A/C include docs, so we'll leave this open to be covered by that story.

PR #1122, opened 2023-09-19
Updating to the DevOps components paradigm

josh-padnick:
This PR updates the DevOps foundations section to use the [components paradigm](https://www.notion.so/gruntwork/DevOps-Components-A-perspective-7f1b07a8ca9f46da84563f3382fe1d84) and creates `TODO:` entries throughout the site to capture the remaining work needed for customers to go live with the next release of the product.

## TODOs

The TODOs are intended to be what we should ideally do by EOD Tuesday, so please fix time and adjust scope accordingly. The key business goal here is to empower customers to set up as much as possible on their own using the docs.

## Adopting the DevOps components paradigm

I made the following high-level changes:

- Add an overview of DevOps foundations, and explains it in terms of components
- Add sidebar entries for each components
- Add overviews for Landing Zone, Network Topology, IaC foundations, and Pipelines components

Specifically, I'm aiming to shift away from the paradigm of "4 elements" and to the paradigm of "semi-independent components." Hopefully over time, our components become increasingly independent. 

This paradigm shift is a change we've been discussing for some time, and my observation is that the components paradigm resonates strongly with multiple customers, though my guess is that we're not all in sync on it yet.

Given that, we should focus on shipping the next useful increment of our product within the time limits we have. If you can work in the remaining docs that customers need to go live as part the paradigm proposed here, please do so. If you cannot for some reason, then please use your judgement to make the call that best serves the customer, and we'll revisit this paradigm later on.


_docs-sources/foundations/terraform/folder_structure.md
```diff
@@ -1,5 +1,14 @@
 # Recommended Folder Structure
 
+TODO: Harmonize this content with:
+- https://github.com/orgs/gruntwork-io/discussions/769
+- https://github.com/orgs/gruntwork-io/discussions/765
+- https://github.com/orgs/gruntwork-io/discussions/756
+
+Please focus on harmonizing the guidance above with the guidance below, subject to our current time constraints. If there are any recommendations that we don't yet implement in our infra-live/infra-modules templates, please create separate stories for them and call them out in Slack so that we can figure out where to prioritize them in future sprints.
```
MoonMoon1919, line 8: Do you have more concrete requests other than harmonizing? Large chunks of this doc was copied from Jim's post.

josh-padnick: I actually didn't review the existing content in this PR in depth. My main thinking was that:

1. We should cover the key items that keep coming up, most of which I believe Jim covers:
   - How to create the backend (e.g. S3 bucket) for storing Terraform state
   - How to structure your folder hierarchy
   - How to handle tagging and labels
   - Whether to use branches per environment
   - How to handle global variables
   - How to handle module default values
1. We should make sure that our recommendations here either reflect Jim's advice, or if we disagree, do so consciously and intentionally (but to be clear, it is ok to disagree!)

josh-padnick:
These changes have all been manually merged in separate PRs, so closing it out.

PR #1148, opened 2023-09-28
chore(CORE-1309): Make docs around limiting actions more explicit

MoonMoon1919:
# What it is

Adds more explicit docs surrounding allow-list for actions for GHE users and docs for GH pro/teams users.

### DIFF
--- a/_docs-sources/pipelines/using-pipelines/index.md
+++ b/_docs-sources/pipelines/using-pipelines/index.md
@@ -63,7 +63,21 @@ You will need to allow the following GitHub Actions for Gruntwork Pipelines to r
 - [dsaltares/fetch-gh-release-asset](https://github.com/dsaltares/fetch-gh-release-asset)
 - [peter-evans/create-pull-request](https://github.com/peter-evans/create-pull-request)
 
-Gruntwork recommends using a tagged version of the `gruntwork-io` actions and a wildcard for the non-Gruntwork actions, as the non-Gruntwork actions are wrapped in the Gruntwork actions. This will require less administrative work to keep policies up to date with versions being used by Gruntwork actions.
+Gruntwork recommends using a tagged version of the `gruntwork-io` actions and a wildcard for the non-Gruntwork actions, as the non-Gruntwork actions are wrapped in the Gruntwork actions. This will require less administrative work to keep policies up to date with versions being used by Gruntwork actions. An example policy can be found below.
+
+```txt
+gruntwork-io/pipelines-dispatch@<newest release>,
+gruntwork-io/pipelines-orchestrate@<newest release>,
+gruntwork-io/pipelines-execute@<newest release>,
+gruntwork-io/terragrunt-action@<newest release>,
+aws-actions/configure-aws-credentials@*,
+Codex-/return-dispatch@*,
+Codex-/await-remote-run@*,
+dsaltares/fetch-gh-release-asset@*,
+peter-evans/create-pull-request@*
+```
+
+Navigate to each Gruntwork repository to retrieve the latest tagged release for each action.
 
 ### GitHub Team and Pro
 
--- a/_docs-sources/pipelines/using-pipelines/index.md
+++ b/_docs-sources/pipelines/using-pipelines/index.md
@@ -63,7 +63,7 @@ You will need to allow the following GitHub Actions for Gruntwork Pipelines to r
 - [dsaltares/fetch-gh-release-asset](https://github.com/dsaltares/fetch-gh-release-asset)
 - [peter-evans/create-pull-request](https://github.com/peter-evans/create-pull-request)
 
-We recommend using a tagged version of the `gruntwork-io` actions and a wildcard for the non-Gruntwork actions, as the non-Gruntwork actions are wrapped in the Gruntwork actions. This will require less administrative work to keep policies up to date with versions being used by Gruntwork actions.
+Gruntwork recommends using a tagged version of the `gruntwork-io` actions and a wildcard for the non-Gruntwork actions, as the non-Gruntwork actions are wrapped in the Gruntwork actions. This will require less administrative work to keep policies up to date with versions being used by Gruntwork actions.
 
 ### GitHub Team and Pro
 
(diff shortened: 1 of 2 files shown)


_docs-sources/pipelines/using-pipelines/index.md
```diff
@@ -44,13 +44,27 @@ The following is an example of the recommended settings for branch protection:
...
+- [gruntwork-io/pipelines-orchestrate](https://github.com/gruntwork-io/pipelines-orchestrate)
+- [gruntwork-io/pipelines-execute](https://github.com/gruntwork-io/pipelines-execute)
+- [gruntwork-io/terragrunt-action](https://github.com/gruntwork-io/terragrunt-action)
+- [aws-actions/configure-aws-credentials](https://github.com/aws-actions/configure-aws-credentials)
+- [Codex-/return-dispatch](https://github.com/Codex-/return-dispatch)
+- [Codex-/await-remote-run](https://github.com/Codex-/await-remote-run)
+- [dsaltares/fetch-gh-release-asset](https://github.com/dsaltares/fetch-gh-release-asset)
+- [peter-evans/create-pull-request](https://github.com/peter-evans/create-pull-request)
+
+Gruntwork recommends using a tagged version of the `gruntwork-io` actions and a wildcard for the non-Gruntwork actions, as the non-Gruntwork actions are wrapped in the Gruntwork actions. This will require less administrative work to keep policies up to date with versions being used by Gruntwork actions.
```
oredavids, line 66: Worth including a copy-pastable example here?

oredavids (approved):

_docs-sources/pipelines/using-pipelines/index.md
MoonMoon1919: I considered it, but since the versions we are recommending will semi-regularly change, i ultimately opted against it so we wouldn't need to keep it up to date in the docs. If you think it would be helpful I'm happy to add it!

oredavids: I think it could be helpful, we can use a placeholder for the version number so its generic.

josh-padnick: +1 on the copy-pastable example. If it's cumbersome for us to maintain and keep this doc up to date, it would also be cumbersome for customers to keep up to date, so that's an issue we should directly address.

> Gruntwork recommends using a tagged version of the `gruntwork-io` actions and a wildcard

Can we explicitly recommend a specific tag they should use? As a side note, I've noticed our configs reference `@v1.X.Y` of an action, but should we just set them to `@v1` so that we can keep publishing updates and customers automatically get the latest?

MoonMoon1919: Added in [b3d5092](https://github.com/gruntwork-io/docs/pull/1148/commits/b3d50920f24c1da69e2e27df60fbb3bc44b75f09)

oredavids (approved):

_docs-sources/pipelines/using-pipelines/index.md
oredavids: Thanks! Approved.

_docs-sources/pipelines/using-pipelines/index.md
```diff
@@ -44,13 +44,27 @@ The following is an example of the recommended settings for branch protection:
...
+You will need to allow the following GitHub Actions for Gruntwork Pipelines to run successfully:
+- [gruntwork-io/pipelines-dispatch](https://github.com/gruntwork-io/pipelines-dispatch)
+- [gruntwork-io/pipelines-orchestrate](https://github.com/gruntwork-io/pipelines-orchestrate)
+- [gruntwork-io/pipelines-execute](https://github.com/gruntwork-io/pipelines-execute)
+- [gruntwork-io/terragrunt-action](https://github.com/gruntwork-io/terragrunt-action)
+- [aws-actions/configure-aws-credentials](https://github.com/aws-actions/configure-aws-credentials)
+- [Codex-/return-dispatch](https://github.com/Codex-/return-dispatch)
+- [Codex-/await-remote-run](https://github.com/Codex-/await-remote-run)
+- [dsaltares/fetch-gh-release-asset](https://github.com/dsaltares/fetch-gh-release-asset)
+- [peter-evans/create-pull-request](https://github.com/peter-evans/create-pull-request)
```
josh-padnick, line 64: Since we recommend a wildcard allow for these GitHub Actions, we are putting a significant amount of security trust in these third parties. If Peter Evans updates his GitHub Action, 100% of our customers will get it, but we will have never reviewed it. Actually, the very fact that we publish this list means we're telling an attacker exactly which dependencies to compromise.

In traditional app dev, we can mitigate that risk with CVE checks and vulnerability scans. Does anything like that exist for GitHub Actions?

As far as next steps here, I've captured this in [CORE-1326](https://gruntwork.atlassian.net/browse/CORE-1326).

josh-padnick (approved):
LGTM! I left a couple small comments.

josh-padnick:
@MoonMoon1919 Make sure collaborate with @ellisonc as he recently updated where the GitHub Enteprise content lives. I recommend going with his organizational scheme and your content.

_docs-sources/pipelines/using-pipelines/index.md
MoonMoon1919: >  If Peter Evans updates his GitHub Action, 100% of our customers will get it, but we will have never reviewed it

We're pinning the versions of each of the actions we use _inside_ of our actions and workflows. So we will _have_ to review any change to any 3rd party action, increment the version in our actions or workflows, then roll that out to customers.

_docs-sources/pipelines/using-pipelines/index.md
MoonMoon1919: >  should we just set them to @v1 so that we can keep publishing updates and customers automatically get the latest

The `@v1.x.x` or `@v1` correspond to tags/releases in the repository. As far as I am aware you have to be explicit about the tag. You can't just say `anything under v1`.

_docs-sources/pipelines/using-pipelines/index.md
josh-padnick: > We're pinning the versions of each of the actions we use inside of our actions and workflows.

Ah, that's a good point. 

> All that being said, ideally using these other actions is temporary while we put more logic into the binary.

Agreed!

_docs-sources/pipelines/using-pipelines/index.md
josh-padnick: Oh got it. If they require @v1.x.x tagging then so be it!

MoonMoon1919:
> @MoonMoon1919 Make sure to collaborate with @ellisonc as he recently updated where the GitHub Enteprise content lives. I recommend going with his organizational scheme and your content.

We had a quick sync offline about how to proceed. I'm going to merge this PR and he'll fold the changes into his PR next.

PR #1286, opened 2023-11-16
Update/Refactor Foundations Guide - CORE-1457

oredavids:
Updating & refactoring minor components of the foundations guide.

Closes [CORE-1457](https://gruntwork.atlassian.net/browse/CORE-1457)

[CORE-1457]: https://gruntwork.atlassian.net/browse/CORE-1457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_docs-sources/foundations/landing-zone/add-aws-account.md
+++ b/_docs-sources/foundations/landing-zone/add-aws-account.md
@@ -31,6 +31,19 @@ Before proceeding, ensure you have:
 
   To initiate the process, you can use the GitHub Actions workflow in your `infrastructure-live` repo by following the below steps. Alternatively, you can choose to manually create the account request file and associated PR by following the steps in the **Account Request File (Manual)** dropdown below.
 
+  You will be required to fill in the following parameters:
+
+  - `account_name`: The name of the account in AWS.
+  - `account_email`: The email address to associate with the account.
+  - `organizational_unit_name`: The name of the Organizational Unit(OU) where the new account will be added. Ensure this exactly matches an existing OU name in your organization, for example, “Security”.
+  - `aws_region`: The AWS region where the Gruntwork account baselines will be deployed. e.g., `us-east-1`.
+  - `org_name_prefix`: The prefix to use for some resources created by Gruntwork modules. For instance, S3 bucket files will be prefixed with this value
+  - `sso_user_first_name`: The first name of the user to create in AWS IAM Identity Center (Successor to AWS Single Sign-On(SSO)).
+  - `sso_user_last_name`: The last name of the user to create in AWS IAM Identity Center.
+  - `sso_user_email`: The email address of the user to create in AWS IAM Identity Center. This user will be able to login to the new account using AWS IAM Identity Center.
+  - `account_baseline_modules_version`: The version of the your account baseline modules to use for the new account. e.g., `v0.0.1`
+  - `requested_by`: The GitHub user ID or email address of the entity requesting the new account, for audit purposes.
+
 <Tabs groupId="request-aws-account">
 <TabItem value="GitHub Action Workflow" label="GitHub Action Workflow" default>
 
@@ -61,17 +74,6 @@ The workflow will automatically create a PR to be reviewed.
   account_baseline_modules_version: <ACCOUNT-BASELINE-MODULES-VERSION>
   requested_by: <GITHUB_USER_ID_OR_EMAIL>
   ```
-
-    - `account_name`: The name of the account in AWS.
-    - `account_email`: The email address to associate with the account.
-    - `organizational_unit_name`: The name of the Organizational Unit(OU) where the new account will be added. Ensure this exactly matches an existing OU name in your organization, for example, “Security”.
-    - `aws_region`: The AWS region where the Gruntwork account baselines will be deployed. e.g., `us-east-1`.
-    - `org_name_prefix`: The prefix to use for some resources created by Gruntwork modules. For instance, S3 bucket files will be prefixed with this value
-    - `sso_user_first_name`: The first name of the user to create in AWS IAM Identity Center (Successor to AWS Single Sign-On(SSO)).
-    - `sso_user_last_name`: The last name of the user to create in AWS IAM Identity Center.
-    - `sso_user_email`: The email address of the user to create in AWS IAM Identity Center. This user will be able to login to the new account using AWS IAM Identity Center.
-    - `account_baseline_modules_version`: The version of the your account baseline modules to use for the new account. e.g., `v0.0.1`
-    - `requested_by`: The GitHub user ID or email address of the entity requesting the new account, for audit purposes.
 
 1. Create a pull request
 
--- a/_docs-sources/foundations/landing-zone/prerequisites.md
+++ b/_docs-sources/foundations/landing-zone/prerequisites.md
@@ -2,8 +2,6 @@
 
 The first step to using Gruntwork Landing Zone is to use AWS Control Tower to create a new multi-account setup, which requires that you meet the prerequisites below.
 
-To use AWS Control Tower you must first enable it, which requires that you have the following:
-
 1. A new AWS Account and a user with administrator permissions. (We recommend using an IAM user with admin permissions rather than the root user)
 
   :::info
--- a/docs/foundations/landing-zone/add-aws-account.md
+++ b/docs/foundations/landing-zone/add-aws-account.md
@@ -31,6 +31,19 @@ Before proceeding, ensure you have:
 
   To initiate the process, you can use the GitHub Actions workflow in your `infrastructure-live` repo by following the below steps. Alternatively, you can choose to manually create the account request file and associated PR by following the steps in the **Account Request File (Manual)** dropdown below.
 
+  You will be required to fill in the following parameters:
+
+  - `account_name`: The name of the account in AWS.
+  - `account_email`: The email address to associate with the account.
+  - `organizational_unit_name`: The name of the Organizational Unit(OU) where the new account will be added. Ensure this exactly matches an existing OU name in your organization, for example, “Security”.
+  - `aws_region`: The AWS region where the Gruntwork account baselines will be deployed. e.g., `us-east-1`.
+  - `org_name_prefix`: The prefix to use for some resources created by Gruntwork modules. For instance, S3 bucket files will be prefixed with this value
+  - `sso_user_first_name`: The first name of the user to create in AWS IAM Identity Center (Successor to AWS Single Sign-On(SSO)).
+  - `sso_user_last_name`: The last name of the user to create in AWS IAM Identity Center.
+  - `sso_user_email`: The email address of the user to create in AWS IAM Identity Center. This user will be able to login to the new account using AWS IAM Identity Center.
+  - `account_baseline_modules_version`: The version of the your account baseline modules to use for the new account. e.g., `v0.0.1`
+  - `requested_by`: The GitHub user ID or email address of the entity requesting the new account, for audit purposes.
+
 <Tabs groupId="request-aws-account">
 <TabItem value="GitHub Action Workflow" label="GitHub Action Workflow" default>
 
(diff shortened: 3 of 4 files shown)


_docs-sources/foundations/landing-zone/enable-control-tower.md
```diff
@@ -33,25 +37,37 @@ waiting on Control Tower Operations at the conclusion of the setup flow.
 ### Review pricing and select Regions
 
 1. Under **Region deny setting** select Enabled
+
    1. This ensures Control Tower policies and controls are unable to be bypassed by using a non-governed region
 
 1. Under **Select additional Regions for governance** select all regions where you plan to operate.
 
+   <details>
```
ebeneliason, line 45: This is a fairly clever solution. I think we might be able to find some sort of plugin, or even write a custom React component ourselves to create an expandable thumbnail, but let's roll with this as is for now!

_docs-sources/foundations/landing-zone/enable-control-tower.md
```diff
@@ -84,38 +114,50 @@ Control Tower Creation will take around an hour to complete
...
 1. Setting up the landing zone can take up to one hour. You will see a notification like the one below with the estimated time it will take for all the resources to be created.
 
-  ![Landing Zone Setup Status](/img/devops-foundations/account/control-tower-setup-status.png)
+   ![Landing Zone Setup Status](/img/devops-foundations/account/control-tower-setup-status.png)
 
-  :::tip
+   :::tip
 
-  You can safely close the browser once you see this.
+   You can safely close the browser once you see this.
```
ebeneliason, line 121: I don't understand this — which browser? My initial read was the one I'm reading the docs in. But maybe you mean the AWS Control Tower console? But then the very next step instructs the user to navigate to the AWS Control Tower dashboard.

_docs-sources/foundations/landing-zone/prerequisites.md
```diff
@@ -16,45 +18,78 @@ To use AWS Control Tower you must first enable it, which requires that you have
...
+
 1. A KMS Key for encrypting Control Tower Resources with a suitable permissions policy (See Guide Below)
+
 1. A home region selection where your Control Tower configuration will reside.
    :::caution
    Your home Region is the AWS Region in which you'll run most of your workloads or store most of your data. It **cannot** be changed after you've set up your AWS Control Tower landing zone. For more information about how to choose a home Region, see [Administrative tips for landing zone setup](https://docs.aws.amazon.com/controltower/latest/userguide/tips-for-admin-setup.html).
    :::
 
-### KMS Key Creation
+---
```
ebeneliason, line 30: This rule confuses me again. Perhaps we're interpreting the structure of the page differently. Then again, if it's wrong, maybe we should address the hierarchy of the content rather than add in rules to divide it arbitrarily. I do think the hierarchy of this page is a little confusing in its own right. Collapsing the images seems like a win, but I'd leave the rules out and we can discuss a better way to structure the content.

For example, there's a whole section for pipelines, which just has one sentence which points to the Pipelines docs. That's actually misleading, because if you follow all the docs to set up a standalone Pipeline you'll be worse off, since that won't support Landing Zone, which requires a multi-account configuration. We only tell people to install Pipelines down in the Initial Setup page of the IaC Foundations section. But that requires having Control Tower already configured. Eek!

This really does need some larger rework. Perhaps we should just remove the Pipelines section from the prerequisites here for now, lest we lead people astray. The Next Steps section doesn't add much either, to be honest.

ebeneliason (commented):
Thanks, Ryan! There are some definite improvements here, but also a few things that confused me. Let me know your thoughts and I'll be happy to give this another look.

_docs-sources/foundations/landing-zone/enable-control-tower.md
```diff
@@ -14,6 +16,8 @@ This Guide should take about 1hr 15min to complete, most of that time will be sp
 waiting on Control Tower Operations at the conclusion of the setup flow.
 :::
 
+---
```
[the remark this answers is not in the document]
arsci: I was just trying to break the page up a bit. I found that while I was going through these guides and flipping back and forth between the console, it was easy to lose my place. Having some visual breaks always helps me so I chose to put a few of these in. In terms of the overall guide I would agree that they are completely unnecessary, but the visual break made it a little easier for me so I gave it a try. Admittedly there's probably a better way to accomplish this, and I'm okay with removing them if you think it's better without!

_docs-sources/foundations/landing-zone/enable-control-tower.md
arsci: I didn't write the original line, but I think the intent here was following up with the previous line that it can take up to an hour, and just advising that you don't need to keep the console page open for an hour while you wait (and if you accidentally do close the tab it is not harmful/won't cancel/abort the operation AWS is performing in the background).

_docs-sources/foundations/landing-zone/prerequisites.md
arsci: Yes I agree larger rework is needed, there are definitely some gaps. 

To address your initial comment on the rule usage, it's similar to the other comment I left in that I was just trying to break the page up for readability. It may not be the right choice here. (and very well may be better addressed by reworking the organization of the guides instead).

_docs-sources/foundations/landing-zone/add-aws-account.md
```diff
@@ -6,46 +6,50 @@ This document provides instructions for provisioning a new AWS account using Gru
...
+
+  Your GitHub org MUST have fine-grained personal access tokens enabled before you can create the access tokens and secrets!
+
+  :::
+
+  :::info
+
+  Instructions for configuring access tokens and secrets can be found in the respective repo Readme files, as well as in [the Machine Users](/pipelines/security/machine-users) guide.
+
+  :::
```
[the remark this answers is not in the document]
arsci: That's a good point. I've removed most of these callouts and left it just as inline text, I do think it looks better. I left the last one as it's important and is something that has tripped us up on multiple occasions. Let me know what you think!

_docs-sources/foundations/landing-zone/add-aws-account.md
```diff
@@ -6,46 +6,50 @@ This document provides instructions for provisioning a new AWS account using Gru
...
-- `sso_user_first_name`: The first name of the user to create in AWS IAM Identity Center (Successor to AWS Single Sign-On(SSO)).
-- `sso_user_last_name`: The last name of the user to create in AWS IAM Identity Center.
-- `sso_user_email`: The email address of the user to create in AWS IAM Identity Center. This user will be able to login to the new account using AWS IAM Identity Center.
-- `account_baseline_modules_version`: The version of the your account baseline modules to use for the new account. e.g., `v0.0.1`
-- `requested_by`: The GitHub user ID or email address of the entity requesting the new account, for audit purposes.
-
-:::note
-
-If you have configured GitHub Actions in your `infrastructure-live` repository with a Gruntwork Landing Zone Account Factory workflow, you can invoke that workflow via the GitHub UI or programmatically. This workflow will automatically create the account request file and open a pull request on your behalf. After doing so, jump to [step 3](#3-review-and-merge-the-account-request-pr).
+To initiate the process, you will use the GitHub Actions workflow in your `infrastructure-live` repo.
```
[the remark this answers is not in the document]
arsci: That's a great point, I don't think we want to remove it entirely. 

I've added the old docs back. I put two drop down tabs, one for the GitHub Action Workflow instructions, which I set to auto-open when the page loads, as well as the original manual Account Request File workflow which will be closed when the page loads.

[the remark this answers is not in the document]
ebeneliason: When we present two alternative approaches like this, I think it would be better to leverage [the built-in tabs feature](https://docusaurus.io/docs/markdown-features/tabs). Everything else about this change looks good.

_docs-sources/foundations/landing-zone/enable-control-tower.md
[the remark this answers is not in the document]
ebeneliason: I can understand the issue of losing one's place. That said, as placed I don't think these align with the content hierarchy particularly well, since they separate the header of one section from the body it pertains to. If anything, it seems like you've "set aside" the pre-requisites section, in a way. I wonder if there's some way we could add a distinct visual treatment to prerequisites _everywhere_ so that they stand out and can be easily identified. That seems out of scope for this change, though…

_docs-sources/foundations/landing-zone/enable-control-tower.md
ebeneliason: Ah, that makes sense. I think that context would be very useful to add to the callout. Here's an attempt:
> You can safely close your browser tab once you see this notice. The setup process will proceed unaffected in the background.

_docs-sources/foundations/landing-zone/prerequisites.md
ebeneliason: Okay, here's a concrete proposal I feel we could do with this change. Let me know your thoughts:
* Remove the "Gruntwork Pipelines" section
* Remove the "Next steps" section
* Move the "Creating the KMS key" section into a details block within item 3 in the list of pre-reqs
* Remove the "Prerequisites for enabling AWS Control Tower" _header_ (it's redundant, since at this point, the only content on the page is the numbered list of 4 pre-reqs)
* Remove all the rules, which should no longer be needed on the much-shortened page

arsci: Great suggestions, this is much more simplified. The only suggestions I wasn't able to incorporate was embedding the entire Creating a KMS key sections into a details block. For whatever reason I was unable to get details blocks to render inside another details block (the screenshots). I think it looks better though, so let me know what you think about how it looks now! I can keep tinkering with it if it's still not quite right.

_docs-sources/foundations/landing-zone/add-aws-account.md
```diff
@@ -6,46 +9,76 @@ This document provides instructions for provisioning a new AWS account using Gru
...
+    - `organizational_unit_name`: The name of the Organizational Unit(OU) where the new account will be added. Ensure this exactly matches an existing OU name in your organization, for example, “Security”.
+    - `aws_region`: The AWS region where the Gruntwork account baselines will be deployed. e.g., `us-east-1`.
+    - `org_name_prefix`: The prefix to use for some resources created by Gruntwork modules. For instance, S3 bucket files will be prefixed with this value
+    - `sso_user_first_name`: The first name of the user to create in AWS IAM Identity Center (Successor to AWS Single Sign-On(SSO)).
+    - `sso_user_last_name`: The last name of the user to create in AWS IAM Identity Center.
+    - `sso_user_email`: The email address of the user to create in AWS IAM Identity Center. This user will be able to login to the new account using AWS IAM Identity Center.
+    - `account_baseline_modules_version`: The version of the your account baseline modules to use for the new account. e.g., `v0.0.1`
+    - `requested_by`: The GitHub user ID or email address of the entity requesting the new account, for audit purposes.
```
ebeneliason, line 74: I recommend pulling this list of parameters out of the tabs themselves since they apply to both approaches.

_docs-sources/foundations/landing-zone/prerequisites.md
```diff
@@ -2,12 +2,6 @@
 
 The first step to using Gruntwork Landing Zone is to use AWS Control Tower to create a new multi-account setup, which requires that you meet the prerequisites below.
 
-## Gruntwork Pipelines
-
-In order to get the full functionality of Gruntwork Landing Zone, you must use [Gruntwork Pipelines](/pipelines/overview/).
-
-## Enabling AWS Control Tower
-
 To use AWS Control Tower you must first enable it, which requires that you have the following:
```
ebeneliason, line 5: I recommend cutting this sentence entirely, as it's now rather redundant with the intro sentence above.

ebeneliason (commented):
We're very close! Just a couple minot comments left here.

_docs-sources/foundations/landing-zone/add-aws-account.md
arsci: Pulled this out and added it before the tabbed sections. Let me know what you think!

ebeneliason: Looks great!

I'm noticing minor nits now that I look more closely, but they are with the original content and I think we've held this PR up long enough! I'll approve it as is and use the "edit this page" feature to make a few typo fixes.

ebeneliason (approved):
LGTM!

PR #1361, opened 2023-12-19
Resolve inconsistency in comments for style guide

KyleKing:
The Style Guide indicate that a single `#` should be used, but then has examples with double `#`, which is confusing

https://github.com/KyleKing/docs/blame/89ba5ba2b1468f55bb77b9b7b1f418eb4af586fe/docs/guides/style/terraform-style-guide.md#L169-L171

Unless the `## --` is the intended guidance, I can also make that change

---

*The changes look correct in deployment: https://6581f385a7bf9a0008663549--pensive-meitner-faaeee.netlify.app/guides/style/terraform-style-guide*

### DIFF
--- a/docs/guides/style/terraform-style-guide.md
+++ b/docs/guides/style/terraform-style-guide.md
@@ -178,27 +178,27 @@ optional variables (with defaults) using block comments.
 Example:
 
 ```hcl
-## ---------------------------------------------------------------------------------------------------------------------
-## ENVIRONMENT VARIABLES
-## Define these secrets as environment variables
-## ---------------------------------------------------------------------------------------------------------------------
+# ---------------------------------------------------------------------------------------------------------------------
+# ENVIRONMENT VARIABLES
+# Define these secrets as environment variables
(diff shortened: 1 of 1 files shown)


KyleKing:
Just following up, does someone have time to take a look at this PR?

KyleKing:
Just following up. This PR makes a minimal change to remove conflicting guidance in the style guide on block comments. Reviewing other code published by Gruntworks supports the change I've submitted here because the examples I found all use the single `# ---` syntax rather than `## ---`.

Examples:

- https://github.com/gruntwork-io/terraform-aws-utilities/blob/a3217852061d0115dbbc2245bdb1abf8f4901574/modules/run-pex-as-data-source/main.tf#L1-L6
- https://github.com/gruntwork-io/terragrunt-infrastructure-modules-example/blob/776c82dc161be3be9303070628056df95ec2b907/asg-elb-service/main.tf#L1-L9

ZachGoldberg (approved):

PR #1683, opened 2024-04-30
feat: Deprecating `infrastructure-pipelines` architecture docs

yhakbar:
Deprecating the legacy documentation for Gruntwork Pipelines that leveraged the `infrastructure-pipelines` repository.


ZachGoldberg (approved):
Went through most of this, made one slight change.  Really great work here Yousif!  I think there's more we can clarify about how access control works, future Pipelines actions, highlighting the enterprise delegated-repo feature etc., but none of that IMO should be blocking for a day 1 launch of these docs.

_docs-sources/pipelines/security/repository-access.md
```diff
@@ -1,9 +1,45 @@
...
+If you are using the older version of Gruntwork Pipelines that includes the `infrastructure-pipelines` repository, click [here](../../infrastructure-pipelines/overview/deprecation.md) to learn more about the deprecation of that version.
+:::
 
-The diagram below visually illustrates the above teams:
+Gruntwork recommends that you grant permissions to GitHub repositories by defining three [GitHub Teams](https://docs.github.com/en/organizations/organizing-members-into-teams/about-teams), which should map to three separate personas in your organization. Each team and its permissions are designed to apply the [_principle of least privilege_](https://en.wikipedia.org/wiki/Principle_of_least_privilege) to each individual (or machine user) in your organization for them to be able to perform changes to your infrastructure.
```
josh-padnick, line 9:
```suggestion
## Recommended roles and GitHub teams

Gruntwork recommends that you grant permissions to GitHub repositories by defining three [GitHub Teams](https://docs.github.com/en/organizations/organizing-members-into-teams/about-teams), which should map to three separate personas in your organization. Each team and its permissions are designed to apply the [_principle of least privilege_](https://en.wikipedia.org/wiki/Principle_of_least_privilege) to each individual (or machine user) in your organization for them to be able to perform changes to your infrastructure.
```

_docs-sources/infrastructure-pipelines/security/branch-protection.md
```diff
@@ -0,0 +1,49 @@
+# Branch Protection
+
+:::info Newer Version Available
+This documentation pertains to an old version of Gruntwork Pipelines which used the `infrastructure-pipelines` repository. [Click here](../../pipelines/overview/) to view documentation for the most recent version.
+:::
+
+Gruntwork Pipelines is designed to be used with a PR based workflow.
```
josh-padnick, line 7:
```suggestion
Gruntwork Pipelines is designed to be used with a pull-request (PR) based workflow.
```

Some users may not be familiar with that acronym. That being said, I think we only need to define it once per page.

_docs-sources/pipelines/security/machine-users.md
```diff
@@ -3,7 +3,13 @@ import TabItem from '@theme/TabItem';
 
 # Machine users
```
josh-padnick, line 4: This content is excellent, but it seems like the closest thing we have to "installation" instructions. Perhaps we should create a new sidebar section called "Initial setup" and list this there? Also, could you clarify when a user would need to do this? Or perhaps we should clarify explicitly that Gruntwork Pipelines is only available with DF, but will be available for standalone installation soon?

josh-padnick (approved):
Yousif, this is great! I left a bunch of comments, but to be clear, feel free to merge whenever you'd like, as we'll wind up iterating on these docs several times.

Overall, the docs do a great job of discovering how Pipelines works under the hood, though, there's very limited info on:

- How to install it as a user or a new team (or whether you can install it outside of DF)
- How to use and configure it as an end user
- What the actual UX is like (it might be worth showing a screenshot so users can imagine the flow without having to set it all up)

For next steps, perhaps you could include what you can from the above, and in a follow-up docs update we can fill in the gaps around usability, configuration and UX.

yhakbar:
> Yousif, this is great! I left a bunch of comments, but to be clear, feel free to merge whenever you'd like, as we'll wind up iterating on these docs several times.
> 
> Overall, the docs do a great job of discovering how Pipelines works under the hood, though, there's very limited info on:
> 
>     * How to install it as a user or a new team (or whether you can install it outside of DF)
> 
>     * How to use and configure it as an end user
> 
>     * What the actual UX is like (it might be worth showing a screenshot so users can imagine the flow without having to set it all up)
> 
> 
> For next steps, perhaps you could include what you can from the above, and in a follow-up docs update we can fill in the gaps around usability, configuration and UX.

Given that nothing seems to require stopping the show, I will merge this, then circle back to your feedback!

PR #1697, opened 2024-05-07
LIB-76: Replace tfenv and tgswitch references with mise

arsci:
### Description

* Replaces tfenv and tgswitch references with mise


gcagle3 (approved):
LGTM!

PR #1699, opened 2024-05-08
fix: WIP - Addressing feedback from #1683

yhakbar:
Addressing comments in #1683

### DIFF
--- a/_docs-sources/foundations/iac-foundations/architecture.md
+++ b/_docs-sources/foundations/iac-foundations/architecture.md
@@ -1,0 +1,33 @@
+# Architecture
+
+:::info Recent Upgrade
+This documentation relates to the latest version of Gruntwork Pipelines released in May 2024.
+
+If you are using the older version of Gruntwork Pipelines that includes the `infrastructure-pipelines` repository, click [here](../../infrastructure-pipelines/iac-foundations/initial-setup.md) to get the documentation for that version.
+:::
+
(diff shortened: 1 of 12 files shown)


josh-padnick:
I decided to review the work in progress and had a few thoughts in no particular order:

- **Where do the architecture components live?** On https://deploy-preview-1699--pensive-meitner-faaeee.netlify.app/pipelines/architecture/, we describe the architecture components, but we don't say where they live, how they're invoked, etc. Perhaps we could make this concrete. For example, "the orchestrator is a Go binary provided by Gruntwork that is invoked by your GitHub Actions workflow after you make a commit or merge." 
- **A picture's worth 1,000 words.** A simple, rough visual diagram showing a basic information flow could go a long way to helping users understand how Pipelines works.
- **Instructions on how to use Pipelines.** I think I mentioned this in my last feedback, but whether as part of this PR or a follow-on one, we could use instructions on how to use Pipelines. For example, you make a commit, and that automatically invokes Pipelines. Ideally, we could include screenshots, though those will all change in a few days once we finish the UX polish.
- **Explicit installation instructions.** We should also have an explicit page on how to install Pipelines, even if it requires DF to do so now and the user can't actually install it themselves.
- **We should work in small batches.** Large PRs are tough to work on, so perhaps do what you can to get this PR over the finish line, and we'll figure out updates to these other items in upcoming PRs.

PR #1720, opened 2024-05-15
fix: More migration guide fixes

yhakbar:


_docs-sources/pipelines/upgrading/upgrading-from-infrastructure-pipelines.md
```diff
@@ -26,6 +26,10 @@ Be extra careful to ensure that you have replaced all instances of `<REPLACE>` i
 
 ## Step 2: Run the `bootstrap.yml` workflow
 
+:::caution
+Make sure that you have the appropriate tokens and secrets configured to allow your workflows to access the necessary secrets to interact with your infrastructure. For more information on how to do this, visit the instructions in [Machine Users](../security/machine-users.md).
```
oredavids, line 30: Given that this is an upgrade guide, I would suggest updating the language to explicitly state the old tokens should be deleted and instructions followed to create new tokens.

oredavids (approved):
LGTM with one suggestion.

_docs-sources/pipelines/upgrading/upgrading-from-infrastructure-pipelines.md
yhakbar: That stuff is at the end so that they don't break `infrastructure-pipelines` until after they have a working version of modern Pipelines:
https://github.com/gruntwork-io/docs/blob/master/_docs-sources/pipelines/upgrading/upgrading-from-infrastructure-pipelines.md#delete-old-tokens-and-secrets

PR #1813, opened 2024-07-02
chore: link to our mise thoughts

ZachGoldberg:


yhakbar (approved):

oredavids (commented):
`regenerate` has not been run hence the docs will not actually be updated. The bot should have helped with the update and we may have to investigate that when we can. A local run should suffice for now.

PR #1825, opened 2024-07-12
feat: Adding docs on HCL configs

yhakbar:


docs/pipelines/upgrading/upgrading-from-infrastructure-pipelines.md
oredavids, line 1: I don't see a corresponding _docs-sources file. Is this fixing a problem or introducing one 😁?

_docs-sources/pipelines/overview/configuration.md
```diff
@@ -1,87 +1,360 @@
...
+In most circumstances, the same role would be assumed by multiple units of configuration within a repository (e.g. all units within a given directory configure resources for the same AWS account). In that circumstance, it would be more appropriate not to define a `gruntwork.hcl` file at all, and instead define a `.gruntwork` directory at the root of the repository with a file ending in `.hcl` that contains configurations that are applicable to all units within that directory.
+
+e.g.
+
+```hcl
+# .gruntwork/environments.hcl
+environment "an_account" {
+  filter {
+    paths = ["an-account/*"]
+  }
```
oredavids, line 51: Environments are not equivalent to accounts right? If yes, we should carefully consider the use of "an-account" here. It may be a reader's first engagement with our "environments" declaration and may form a mental model that environments **are** accounts.

_docs-sources/pipelines/overview/configuration.md
```diff
@@ -1,87 +1,360 @@
...
+The label applied to an environment block is the name of the environment. This is a user-defined label for the environment, and must be globally unique.
+
+e.g.
+
+```hcl
+# .gruntwork/environments.hcl
+environment "an_environment" {
+  filter {
+    paths = ["an-environment/*"]
+  }
```
oredavids, line 92: Similar comment on the paths example here, I would suggest paths that are not the same as the environment name.

_docs-sources/pipelines/overview/configuration.md
```diff
@@ -1,87 +1,360 @@
...
+      authentication_profile = {
+        plan_iam_role  = "role-to-assume-for-plans"
+        apply_iam_role = "role-to-assume-for-applies"
+      }
+    }
+  }
+}
 ```
+
+In this example, the `an_environment` environment is defined to match all units located within the `an-environment` directory sibling to the `.gruntwork` directory. All units that match this filter will assume the `role-to-assume-for-plans` role in the AWS account with ID `an-aws-account-id` when running Terragrunt plan commands by Pipelines.
```
oredavids, line 106: If my path suggestion above is taken, then the reference here should also be changed.

_docs-sources/pipelines/overview/configuration.md
```diff
@@ -1,87 +1,360 @@
...
+```hcl
+# .gruntwork/aws_accounts.hcl
+aws_accounts "all" {
+  path = "aws/accounts.yml"
+}
+```
+
+In this example, the `all` AWS accounts block is defined in a file named `aws_accounts.yml` within the `.gruntwork` directory. The `all` AWS accounts block references an external file located at `aws/accounts.yml` that contains the definitions of AWS accounts.
+
+DevOps Foundations customers may be familiar with the `accounts.yml` file as a file that is used by Account Factory to define the configurations of AWS accounts. In an effort to reduce the burden of migrating to the new configuration system, Pipelines uses the same schema for the `accounts.yml` file as Account Factory. This means that the `accounts.yml` file that is used by Account Factory can be used by the `aws_accounts` block without modification.
```
oredavids, line 207:
```suggestion
DevOps Foundations customers may be familiar with the `accounts.yml` file as a file that is used by Account Factory to define the configurations of AWS accounts. This means that the `accounts.yml` file that is used by Account Factory can be used by the `aws_accounts` block without modification.
```
I do not think it is necessary to add that extra info since we also say portability was a factor later on. I would also suggest moving this section below the schema so that it has already been introduced.

_docs-sources/pipelines/overview/configuration.md
```diff
@@ -1,87 +1,360 @@
...
+  id: "an-aws-account-id"
+  # optional: The email address of the account owner
+  owner_email: "an-email-address"
+  # optional: Whether or not a VPC has been created in the account. Default is false.
+  vpc_created: true
+```
+
+Note that multiple AWS Accounts blocks can be defined, pointing to different `accounts.yml` files. This allows for the segmentation of AWS accounts into different YAML files for organizational purposes.
+
+The decision to leverage YAML files instead of HCL files for defining the configurations for AWS accounts was an intentional decision to increase the portability of these configurations for usage outside of Pipelines. Tools like [Terragrunt](https://github.com/gruntwork-io/terragrunt/) and [yq](https://github.com/mikefarah/yq) can be used to leverage these files, as they are more portable than HCL files.
```
oredavids, line 224:
```suggestion
:::info
The decision to leverage YAML files instead of HCL files for defining the configurations for AWS accounts was an intentional decision to increase the portability of these configurations for usage outside of Pipelines. Tools like [Terragrunt](https://github.com/gruntwork-io/terragrunt/) and [yq](https://github.com/mikefarah/yq) can be used to leverage these files, as they are more portable than HCL files.
:::
```

_docs-sources/pipelines/overview/configuration.md
```diff
@@ -1,87 +1,360 @@
...
+
+*Supported Attributes:*
```
oredavids, line 226: Best in a separate PR; we could consider similar [styling](https://deploy-preview-1825--pensive-meitner-faaeee.netlify.app/reference/services/app-orchestration/amazon-ecs-cluster#required) we use on module inputs/outputs with the added benefit of showing up in the table of contents.

_docs-sources/pipelines/upgrading/upgrading-from-config-yml.md
```diff
@@ -0,0 +1,106 @@
+# Upgrading from config.yml
+
+Previously, certain configurations for Pipelines were defined within a `.gruntwork/config.yml` file located at the root of repositories using Pipelines.
+
+To improve the quality of the Pipelines experience, configurations are now defined in the Pipelines Configuration as Code system. This system allows for more flexibility and control over configurations, allowing Pipelines to be used in a wider variety of use-cases, for a larger number of users.
```
ZachGoldberg, line 5:
```suggestion
To improve the extensibility of the Pipelines experience, configurations are now defined in the Pipelines Configuration as Code system. This system allows for more flexibility and control over configurations, allowing Pipelines to be used in a wider variety of use-cases, for a larger number of users.
```

_docs-sources/pipelines/upgrading/upgrading-from-config-yml.md
```diff
@@ -0,0 +1,106 @@
...
+This means that if you have a `.gruntwork/config.yml` file in your repository, you can continue to use it as you have been, and Pipelines will continue to work as expected. If you do introduce any HCL configurations into your `.gruntwork` directory or introduce a `gruntwork.hcl` file into a directory, Pipelines will begin to use the HCL configuration system instead of the YAML configuration system.
+:::
+
+## Should I Migrate?
+
+At this time, there is no need to migrate from YAML configurations to HCL configurations. Pipelines will continue to work as expected with YAML configurations.
+
+Gruntwork does not currently recommend migrating from YAML configurations to HCL configurations. There may be a future point where YAML configurations are deprecated, but there will be robust tooling and documentation to support this migration when it becomes necessary.
+
+## YAML Configurations
```
ZachGoldberg, line 19: It's a bit misleading to me that the docs for the YAML config live in a page called "Upgrading from Yml" -- IMO it should still live as its own first class page during this transition period

_docs-sources/pipelines/upgrading/upgrading-from-config-yml.md
```diff
@@ -0,0 +1,106 @@
...
+  # Version of the https://github.com/gruntwork-io/terraform-aws-security modules used in templates
+  module-security-version: v0.68.3
+  # Version of the https://github.com/gruntwork-io/terraform-aws-control-tower modules used in templates
+  control-tower-modules-version: v0.4.2
+  # Version of the https://github.com/gruntwork-io/terraform-aws-architecture-catalog modules used in templates
+  arch-catalog-version: v2.2.5
+  # (Optional) Override the repository in use for the architecture catalog
+  arch-catalog-repo-name: gruntwork-io/terraform-aws-architecture-catalog
+  # (Optional) Override the base path within the architecture catalog
+  arch-catalog-base-path: ./terraform-aws-architecture-catalog
```
ZachGoldberg, line 42: Also, I added like 6 new config options today :sweat:

docs/pipelines/overview/configuration.md
```diff
@@ -1,95 +1,368 @@
...
+:::note
+This page includes documentation for HCL Pipelines Configuration as Code.
 
-:::warning This page is under active development and is a work in progress, June 2024
+If you are using an older version of Pipelines, please refer to the documentation on [Upgrading from config.yml](../upgrading/upgrading-from-config-yml.md).
 :::
 
-Pipelines behavior is primarily configured via a yaml file at `.gruntwork/config.yml` in your repository. Pipelines also reads [.mise.toml](https://docs.gruntwork.io/pipelines/upgrading/upgrading-from-infrastructure-pipelines#adding-misetoml) at the root of the repository to determine what version of OpenTofu/Terraform and Terragrunt to use.
+:::caution
+As of July, 2024, the YAML configuration is still the primary mechanism for configuring Pipelines in DevOps Foundations. The best experience in DevOps Foundations remains with the YAML configuration system, as interactions with Account Factory still need improvement in the HCL configuration system.
```
ZachGoldberg, line 10: As I'm reading this, maybe it makes more sense to more away from the notion that HCL is primary, not even reference migration, and instead brand the HCL configs "experimental" or "beta" or some such, and frame it the opposite way - this HCL config is specifically for brownfield support, and is not intended for all customers yet.

docs/pipelines/overview/configuration.md
```diff
@@ -1,95 +1,368 @@
...
+```
+
+Placing this configuration in a `gruntwork.hcl` file in the same directory as a `terragrunt.hcl` file will cause Pipelines to assume the `role-to-assume-for-plans` role in the AWS account with ID `an-aws-account-id` when running Terragrunt plan commands, using [OIDC](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services) to authenticate to AWS and assume that role.
+
+In most circumstances, the same role would be assumed by multiple units of configuration within a repository (e.g. all units within a given directory configure resources for the same AWS account). In that circumstance, it would be more appropriate not to define a `gruntwork.hcl` file at all, and instead define a `.gruntwork` directory at the root of the repository with a file ending in `.hcl` that contains configurations that are applicable to all units within that directory.
+
+e.g.
+
+```hcl
+# .gruntwork/environments.hcl
```
ZachGoldberg, line 47: I like that you introduce `gruntwork.hcl` in a single folder first, then walk through and expand to the `filter` usecase in teh root of the repo, nice.

docs/pipelines/overview/configuration.md
```diff
@@ -1,95 +1,368 @@
...
+
+In this example, all the units located within the `an-account` directory sibling to the `.gruntwork` directory will assume the `role-to-assume-for-plans` role in the AWS account with ID `an-aws-account-id` when running Terragrunt plan commands by Pipelines.
+
+A typical approach to building Pipelines configurations is to first define minimal configurations that address the most common use-cases, and then to refactor and generalize those configurations as needed to reduce repetition.
+
+More details on how these configurations are defined will be detailed below.
+
+## Global Configurations
+
+The configurations found within a `.gruntwork` directory found in the current working directory, or a parent directory of the current working directory are referred to as global configurations. These configurations are typically applicable within a wide range of contexts within a repository, and are the primary mechanism for configuring Pipelines.
```
ZachGoldberg, line 73:
```suggestion
The configurations found within either a `.gruntwork` directory, the current working directory, or a parent directory of the current working directory are referred to as global configurations. These configurations are typically applicable within a wide range of contexts within a repository, and are the primary mechanism for configuring Pipelines.
```

docs/pipelines/overview/configuration.md
```diff
@@ -1,95 +1,368 @@
...
+
+Pipelines will attempt to find exactly one directory named `.gruntwork` when it is attempting to discover configurations. It will not continue to search for configurations in parent directories once it finds a `.gruntwork` directory.
```
ZachGoldberg, line 75: This is a smidge complicated, maybe worth a visual or an example

docs/pipelines/overview/configuration.md
```diff
@@ -1,95 +1,368 @@
...
+
+Note that you will frequently see filenames for configurations within the `.gruntwork` directory that are named after the configuration block that they define. This is a common pattern, but not a requirement. Any `.hcl` file found within a `.gruntwork` directory will be parsed by Pipelines as one global configuration.
+
+### Environment Blocks
+
+Environment blocks are used to define configurations that are applicable to a specific environment within a repository.
+
+The label applied to an environment block is the name of the environment. This is a user-defined label for the environment, and must be globally unique.
```
ZachGoldberg, line 83: We haven't introduced the concept of labels yet, this seems confusing to me.  "How do I use a label?  The word label isn't in the sample below!"

docs/pipelines/overview/configuration.md
```diff
@@ -1,95 +1,368 @@
...
+        apply_iam_role = "role-to-assume-for-applies"
+      }
+    }
+  }
+}
+```
+
+In this example, the `an_environment` environment is defined to match all units located within the `an-environment` directory sibling to the `.gruntwork` directory. All units that match this filter will assume the `role-to-assume-for-plans` role in the AWS account with ID `an-aws-account-id` when running Terragrunt plan commands by Pipelines.
+
+The environment block is most idiomatically used in conjunction with other configuration blocks, as the contents frequently involve configurations that are applicable to multiple units within a repository.
```
ZachGoldberg, line 108: The wording of this sentence seems unnecessarily complex.  "Environments should reference other blocks rather than redefine configuration everytime."

docs/pipelines/overview/configuration.md
```diff
@@ -1,95 +1,368 @@
...
+
+More on configuration components like [aws_accounts](#aws-accounts-blocks) and [authentication_profile](#authentication-profile-blocks) can be found below.
+
+*Supported Blocks:*
+
+- `filter` (Required): A filter block that determines which units the environment is applicable to. See [Filter Blocks](#filter-blocks) for more information.
+- `authentication` (Required): An authentication block that determines how Pipelines will authenticate with cloud platforms when running Terragrunt commands. See [Authentication Blocks](#authentication-blocks) for more information.
+
+:::caution
+Every unit must be uniquely identified by a single environment block. If a unit is matched by multiple environment blocks, Pipelines will throw an error.
```
ZachGoldberg, line 145: I think this could be clearer if we used the filter/match verbs rather than the verb "identified" -- this should point users directly to the `filter` block.  Maybe also link to how to confirm that environment filters are setup correctly via local usage of pipelines cli.

docs/pipelines/overview/configuration.md
```diff
@@ -1,95 +1,368 @@
...
+:::
+
+### Authentication Profile Blocks
+
+Authentication Profile blocks are configurations used by [authentication](#authentication-blocks) blocks to have commonly re-used authentication configurations codified and referenced by multiple authentication blocks.
+
+Note that this is a configuration block that is qualified by the first label used on it. In the example below, the first label is `aws_oidc`, so the block is an `aws_oidc` authentication profile block.
+
+At this time, the only supported label that can be set for the first label is `aws_oidc`.
```
ZachGoldberg, line 154: There's that word label again - is this used with the same meaning as previously?

docs/pipelines/overview/configuration.md
```diff
@@ -1,95 +1,368 @@
...
+    paths = ["*dev/*"]
+  }
+
+  labels = {
+    "environment" = "dev"
+  }
+}
+```
+
+In this example, the `an_annotation` annotation is defined to match all units located within a directory that contains `dev` at the end of the path. All units that match this filter will have the label `key` set to `value`.
```
ZachGoldberg, line 268:
```suggestion
In this example, the `dev` annotation is defined to match all units located within a directory that contains `dev` at the end of the path. All units that match this filter will have the label `environment` set to `dev`.
```

Also, isn't `environment = dev` redundant with the `environment` block?

docs/pipelines/overview/configuration.md
```diff
@@ -1,95 +1,368 @@
...
+
+*Supported Blocks:*
+
+- `filter` (Required): A filter block that determines which units the annotation is applicable to. See [Filter Blocks](#filter-blocks) for more information.
+- `labels` (Required): A map of key-value pairs that provide metadata about the units that the annotation is applicable to.
```
ZachGoldberg, line 273: Might be worthwhile pointing out that labels aren't used by pipelines today, but will be in the future

ZachGoldberg (changes requested):
Adding request changes so we can align the page titles / big picture organization.  The actual content IMO is close enough to launch and iterate, just wanna nail the URLs / navigation.

docs/pipelines/upgrading/upgrading-from-infrastructure-pipelines.md
yhakbar: Fixing, I believe!

_docs-sources/pipelines/overview/configuration.md
```diff
@@ -1,87 +1,360 @@
...
 
-A sample `config.yml` with a full set of options is below.
+If you are currently using YAML configurations, you should not migrate to the new configuration system at this time.
+:::
+
+Pipelines relies on configurations written in [HashiCorp Configuration Language (HCL)](https://github.com/hashicorp/hcl) to drive dynamic behavior. These configurations are primarily used by Pipelines to determine how to interact with cloud environments within the context of the Infrastructure As Code (IAC) within a code repository.
+
+At a high level, Pipelines will read these configurations by parsing all files that end with `.hcl` within a directory named `.gruntwork` or a single file named `gruntwork.hcl`. In typical usage, the configurations that are global to the IAC in a repository will be defined within a `.gruntwork` directory at the root of the repository, and configurations that are specific to a particular `terragrunt.hcl` file (a unit) located in the same directory as the `terragrunt.hcl` file.
```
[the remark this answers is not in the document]
yhakbar: How strongly do you feel about this? This wording isn't better, IMO.

_docs-sources/pipelines/overview/configuration.md
yhakbar: What would you prefer it be called? I think most folks _should_ name the directories for their environments to be very similar to the names of their environments.

_docs-sources/pipelines/upgrading/upgrading-from-config-yml.md
yhakbar: Do you mind adding them in a follow-up PR?

docs/pipelines/overview/configuration.md
yhakbar: Moved around stuff to make it work more like that.

docs/pipelines/overview/configuration.md
yhakbar: Deferring on this for now.

docs/pipelines/overview/configuration.md
yhakbar: Added an HCL primer before getting into the meat and potatoes.

_docs-sources/pipelines/overview/configuration.md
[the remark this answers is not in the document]
oredavids: I just think it reads better for a long sentence.

_docs-sources/pipelines/overview/configuration.md
oredavids: Thats a good point.

oredavids (approved):

docs/pipelines/upgrading/upgrading-from-infrastructure-pipelines.md
```diff
@@ -14,7 +14,14 @@ Before you begin the migration process, ensure that you have the following prere
 - AWS Administrator (or similar) permissions in all AWS accounts managed by your `infrastructure-live` repository.
 - A solid understanding of Infrastructure as Code (IAC) fundamentals, or support from Gruntwork.
 
-## Step 1: Update the `bootstrap.yml` workflow
+## Step 1: Ensure Machine Users are configured
```
ZachGoldberg, line 17: Is this a rebase/merge oops?

yhakbar: No, it should be included. I previously updated the `_docs-sources` entry for this file in the web UI before without running the server locally, so the corresponding `docs` entry didn't get updated.

The change here is the result of feedback from you that users need to have their machine users configured before they attempt to bootstrap, as the token required to create the pull request won't be available otherwise.

ZachGoldberg: Oh, how weird -- I could _swear_ I reordered those steps already.  Thanks for fixing.

ZachGoldberg (approved):

josh-padnick (commented):
@yhakbar I left a bunch of comments, all but one of which are stylistic nits, and one that tries to really push on non-root `.gruntwork` folders as a concept. Please note that I did not review any docs from "Authentication Profile Blocks" on down.

I realize I'm being a stickler on things, and I appreciate your working with me. If you have a different take, please don't be shy about sharing. Philosophically, we should prioritize whatever best serves the company.

Overall, these docs look great. I'm available most of Wednesday if any live syncing is needed.

_docs-sources/pipelines/overview/configurations-as-code.md
```diff
@@ -0,0 +1,390 @@
...
+
+The best experience in DevOps Foundations remains with the YAML configuration system, as interactions with Account Factory are still under development.
+
+If you are currently using YAML configurations, you should not migrate to the new configuration system at this time.
+:::
+
+Pipelines relies on configurations written in [HashiCorp Configuration Language (HCL)](https://github.com/hashicorp/hcl) to drive dynamic behavior. These configurations are primarily used by Pipelines to determine how to interact with cloud environments within the context of the Infrastructure As Code (IAC) within a code repository.
+
+At a high level, Pipelines will read these configurations by parsing all files that end with `.hcl` within a directory named `.gruntwork` or a single file named `gruntwork.hcl`. In typical usage, the configurations that are global to the IAC in a repository will be defined within a `.gruntwork` directory at the root of the repository, and configurations that are specific to a particular `terragrunt.hcl` file (a unit) located in the same directory as the `terragrunt.hcl` file.
```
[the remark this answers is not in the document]
yhakbar: The `gruntwork.hcl` file must be located next to a `terragrunt.hcl` file if it is to be used, and the `.gruntwork` directory can be in any parent directory (not necessarily at the `.git` root).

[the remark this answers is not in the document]
yhakbar: I think putting a slash before `.gruntwork` is confusing, so I'll avoid pulling that in unless you feel strongly about it.

_docs-sources/pipelines/overview/configurations-as-code.md
```diff
@@ -0,0 +1,390 @@
...
+      authentication_profile = {
+        plan_iam_role = "role-to-assume-for-plans"
+        apply_iam_role = "role-to-assume-for-applies"
+      }
+    }
+  }
+}
+```
+
+In this example, all the units located within the `an-environment` directory sibling to the `.gruntwork` directory will assume the `role-to-assume-for-plans` role in the AWS account with ID `an-aws-account-id` when running Terragrunt plan commands by Pipelines.
```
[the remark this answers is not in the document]
yhakbar: I think it might actually be better to avoid that, because `dev` is more general than what our environments actually relate to. If you have two products, each in their own AWS accounts, and they each have their own dev, stage and prod accounts as part of their SDLC lifecycle, you wouldn't want to think it's appropriate to have a `dev` environment.

You would actually want `product1_dev` and `product2_dev` environments.

_docs-sources/pipelines/overview/configurations-as-code.md
```diff
@@ -0,0 +1,390 @@
...
+  :::
+
+## Global Configurations
+
+Any configurations located within a `.gruntwork` directory either in the current working directory, or a parent directory of the current working directory are referred to as global configurations. These configurations are typically applicable within a wide range of contexts within a repository, and are the primary mechanism for configuring Pipelines.
+
+Pipelines will attempt to find exactly one directory named `.gruntwork` when it is attempting to discover configurations. It will not continue to search for configurations in parent directories once it finds a `.gruntwork` directory.
+
+Note that you will frequently see filenames for configurations within the `.gruntwork` directory that are named after the configuration block that they define. This is a common pattern, but not a requirement. Any `.hcl` file found within a `.gruntwork` directory will be parsed by Pipelines as one global configuration.
```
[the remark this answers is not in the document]
yhakbar: There isn't a hierarchy of configurations, there can only be (up to) one `.gruntwork` directory read by the configuration parser and only (up to) one `gruntwork.hcl` file that will be read by the configuration parser.

The reason we don't require that the `.gruntwork` directory exists at the root of the git repository is that:

1. This supports arbitrary folder structure. You might have all your Terragrunt code in a `terragrunt` directory, in which case it would be nice to have the `.gruntwork` directory nested in there so that your paths are all relative to your `terragrunt` directory. 
2. It works outside of git. This is highly convenient for `pipelines` tests, as we don't have to create an extra `.git` directory in order to have the `.gruntwork` directory detected within a `tests` folder. It also makes it so that we don't have to use special logic for handling recursive traversal outside of `.terragrunt-cache` directories, as cloned OpenTofu modules have their own git directories, and we have to make sure that Pipelines doesn't accidentally think of that directory as the root of configuration.

PR #1829, opened 2024-07-15
Add latest Github discussions and changes

oredavids:
Generated with `docs-sourcer`


ZachGoldberg:
Oh, I had no idea we had a docs sourcer to bring discussions in.  I don't see these actually on the docs site though, how do we find them?

oredavids:
@ZachGoldberg I also just remembered we had this. There are a few shortcomings though; we do not automatically pull in every new discussion largely due to how in the early days of creating the discussion there could be a lot of back and forth and we do not want to spam with several PRs. A monthly or weekly schedule might be best for this so created [this ticke](https://linear.app/gruntwork/issue/DEV-416/run-the-docs-sourcer-knowledge-base-plugin-on-a-schedule)t to tract that.

They are accessible on the docs site only directly with links e.g https://docs.gruntwork.io/discussions/knowledge-base/226. They are meant to be discoverable using the search functionality on the docs site. e.g if you type "Recommended approach to updating terraform" in the bar the first result under the knowledge base section should get you to the same discussion linked earlier.

That search index is custom and each plugin in docs-sourcer can [define](https://github.com/gruntwork-io/docs-sourcer/blob/cb7796a10668587f9884c89cdadb3554f16c6d0c/src/plugins/local-copier/docs-indexing-plugin.ts#L37) how its search is created. Most were written as first pass na d have not been very well tested to see if they are as accurate as they should be so results may not always be what you expect.

The docs site invokes the index command on docs releases [here](https://github.com/gruntwork-io/docs/blob/0462b2e1f0e6a67913b993cd602f000b7596e40a/.circleci/config.yml#L103-L104)

ZachGoldberg (approved):

PR #1911, opened 2024-08-21
DEV-504 ACCESS_CONTROL_WRITE_TOKEN

Resonance1584:


static/img/pipelines/security/ACCESS_CONTROL_WRITE_TOKEN.png
Resonance1584, line 1: Should I include Workflows Read and Write with this token? INFRA_ROOT_WRITE_TOKEN does, which is where this screenshot is copied from. Is this a case of include it because we might need it later, or leave it out because we don't strictly need it?

PR #2093, opened 2024-10-24
Docs Project Milestone 1 - Refactor Navigation

ZachGoldberg:
[Phase 1 of the Docs Update](https://www.notion.so/gruntwork/2024-Docs-Site-Revamp-126d8759a68280c2a12fed43684a4b73)

This is a refactor of the navigation and establishing redirects for old URLs as well as a handful of "minimum viable" new content for the docs site to make sense in the new navigation paradigm.  More content to come.


josh-padnick:
Thanks for posting so quickly! Here are some initial thoughts:

1. **Author experience.** This repo takes a good 40 seconds to get Docusaurus running, and several seconds to refresh any page update. I do think one of our goals for this project should be to slim down our Docusaurus build so that it's a good editor experience. This was one of the main motivations for starting with https://github.com/gruntwork-io/docs-v2. 
2. **Getting started.** The issue I have with the current getting started section is that it's not actually a sequence of steps to get started with Gruntwork, but a collection of "getting started" content. Contrast that with https://developers.hubspot.com/beta-docs/getting-started/overview, which is an actual walkthrough.
3. **Drift detection.** Where did we ultimately decide drift detection should live? Getting started or guides?
4. **We should remove tools and courses.** Tools is outdated. Courses is outdated.
5. **DevOps Foundations.** Where do we plan to put all the DevOps Foundations (e.g. Reference Architecture) content?

Overall, I'm wondering if it would be a better UX if each "domain" is a product, and we're thoughtful about how we organize the other nav bar links. That is, here's what it would look like if each product were a top-level nav item:

image

This does look a little crowded, however, if you look at sites like https://docs.stripe.com/, you'll see that they actually have two "rows" in the nav. In Stripe's case:

Row 1: Logo, Search, Create Account, Sign In
Row 2: 
- Left: Primary Nav
- Right: Reference, Help

Given that, what if we we were to do something similar?

Row 1: Logo, Search, Sign In
Row 2: 
- Left: Get started, DevOps Foundations, IaC Library, Pipelines, Account Factory, Patcher, Catalog
- Right: Reference, Library Catalog, Help (Dropdown)
   - Help: Get Support, Knowledge Base

ZachGoldberg:
This PR, and our first milestone, and the first deliverable we agreed upon yesterday, is just about refactoring nav.  The DX improvements come later.  As an aside, I disabled docs-sourcer locally (its a one line change) and my docs site now boots in about 300 milliseconds.  Definitely don't need a new repo.

We haven't really fleshed out the getting started page yet, so I think it's too soon to really judge. 

I'll also reiterate my strong opposition to putting all the products in the top nav per our conversation yesterday.

PR #2114, opened 2024-10-30
DEV-601 Refactor pipelines config

Resonance1584:


src/components/ConfigOptionItem.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react"
+
+type ConfigOptionItemProps = {
+    Required: boolean
+    FullPath: string
+    Description: string
+    Example?: string
+}
+
+export const ConfigOptionItem: React.FunctionComponent<ConfigOptionItemProps> = ({
```
ZachGoldberg, line 10: Based on your emoji in slack I think you already saw this, but we have an existing pattern of `HclListItem` blocks.  That said, you're documenting yaml here, not HCL, so maybe a new component still makes sense?  Just want to make sure you at least considered the existing components as options

ZachGoldberg:
FYI just in case - nearly all of the config in `config.yaml` is for account factory.  Only the following make sense / have any behavior implications outside of account vending
* deploy-branch-name
* github-org
* tf-binary
* enable-terragrunt-provider-cache
* session-duration
* consolidate-added-or-changed
* consolidate-deleted
* env

docs/2.0/reference/accountfactory/configurations.md
```diff
@@ -0,0 +1,538 @@
...
+
+```yaml
+pipelines:
+  security-account-name: security
+```
+
+</HclListItemExample>
+</HclListItem>
+
+### session-duration
```
Resonance1584, line 338: FIXME: Move back to pipelines side

PR #2117, opened 2024-10-31
DEV-602 - Pipelines auth overview page

ZachGoldberg:


oredavids (approved):

Resonance1584:
This sounds too positive towards tokens IMO. We should treat them as second class and discourage usage _except_ as a fallback for bare minimum plan/apply.

ZachGoldberg:
> This sounds too positive towards tokens IMO. We should treat them as second class and discourage usage _except_ as a fallback for bare minimum plan/apply.

Thanks for the feedback, I agree --  updated the copy on both pages to be a bit stronger in favor of the app.

PR #2199, opened 2024-11-18
remove docs around patcher report being interactive

ceschae:
## description

we're thisclose to not supporting `patcher report --non-interactive=false`. we should just remove the docs so folks don't get confused. that work is tracked [here](https://linear.app/gruntwork/issue/PAT-21/[report]-report-command-always-outputs-json)

there's potential here about examples for how to use `patcher report`'s JSON output to track dependency management within a codebase, but that's not the purpose of this PR.

### DIFF
--- a/docs/2.0/docs/patcher/guides/report.md
+++ b/docs/2.0/docs/patcher/guides/report.md
@@ -1,36 +1,8 @@
 # Patcher Report
 
-Starting in `0.4.2`, Patcher supports two modes: **interactive mode** and **non-interactive mode**.
-
-In interactive mode, the user can browse the discovered module dependencies.
+Patcher supports **non-interactive mode** for `patcher report`.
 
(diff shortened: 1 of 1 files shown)


docs/2.0/docs/patcher/guides/report.md
```diff
@@ -1,37 +1,9 @@
 # Patcher Report
 
-Starting in `0.4.2`, Patcher supports two modes: **interactive mode** and **non-interactive mode**.
-
-In interactive mode, the user can browse the discovered module dependencies.
+Patcher supports **non-interactive mode** for `patcher report`.
```
ZachGoldberg, line 3: This is a bit of a funny way to phrase things - most CLI tools are non-interactive, so saying "non-interactive mode" is like saying "Patcher CLI is a CLI."

ZachGoldberg (approved):

PR #2223, opened 2024-11-25
Update styles to reflect our brand refresh

ebeneliason:
There may be more to optimize, but this provides a reasonably comprehensive first pass we can iterate from as needed. One particular point to consider is the possibility of defining custom prism themes for our code blocks to match those we show on the website. At the moment I've just chosen some pre-defined themes that suit reasonably well given our light and dark color schemes.


ZachGoldberg (approved):
LGTM

PR #2246, opened 2024-12-12
Add hubspot to docs site

odgrim:
- Adds hubspot tracking code following the patterns we use for posthog


ZachGoldberg (changes requested):
Do we need hubspot directly on the page?  Can we do this via a segment integration instead?

PR #2313, opened 2025-01-11
Reference edify : 3 files

EdifyContent:
Consists of only the 3 Reference  files needing changes

### DIFF
--- a/docs/2.0/reference/pipelines/configurations-as-code/index.md
+++ b/docs/2.0/reference/pipelines/configurations-as-code/index.md
@@ -1,19 +1,20 @@
 # Pipelines Configurations as Code
 
-Pipelines relies on configurations written in [HashiCorp Configuration Language (HCL)](https://github.com/hashicorp/hcl) to drive dynamic behavior. These configurations are primarily used by Pipelines to determine how to interact with cloud environments within the context of the Infrastructure As Code (IaC) within a code repository.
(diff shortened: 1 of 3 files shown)


docs/2.0/reference/pipelines/configurations-as-code/index.md
```diff
@@ -28,9 +29,10 @@ unit {
...
+
 
-In most circumstances, the same role would be assumed by multiple Terragrunt units of configuration within a repository (e.g. all units within a given directory configure resources for the same AWS account). In this situation, it would be more convenient to set the AWS authentication at the environment level by declaring an `environment` block in one of the `.hcl` files in the `.gruntwork` directory at the root of the repository, and declaring the AWS authentication configuration there.
+Multiple Terragrunt units within a repository will assume the same AWS role. For instance, all units within a specific directory may provision resources for the same AWS account. Configuring AWS authentication at the environment level is more efficient in these cases. You can do this by defining an `environment` block within one of the `.hcl` files in the `.gruntwork` directory at the repository root and specifying the AWS authentication configuration.
```
ZachGoldberg, line 35:
```suggestion
It is common for multiple Terragrunt units within a repository to assume the same AWS role. For instance, all units within a specific directory may provision resources for the same AWS account. Configuring AWS authentication at the environment level is more efficient in these cases. You can do this by defining an `environment` block within one of the `.hcl` files in the `.gruntwork` directory at the repository root and specifying the AWS authentication configuration.
```

PR #2408, opened 2025-02-10
WIP - File dependency docs

ZachGoldberg:


docs/2.0/docs/pipelines/guides/file-dependencies.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Configuration
+
+Pipelines will automatically detect, based on git commit history, when data files have changed and emit a `FileChanged` event. This will create a new job in your workflow that will invoke Terragrunt with the [`queue-include-units-reading`](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-queue-include-units-reading) option set to the files that have changed. Terragrunt will then scan your code for units that depend on the changed files and add them to the plan/apply.
+
+Note that Terragrunt will only be able to determine which units read the data file if that file is read using a built-in function such as `read_terragrunt_config`. You may want to use the [`mark_as_read`](https://terragrunt.gruntwork.io/docs/reference/built-in-functions/#mark_as_read) function to explicitly ensure that Terragrunt knows to include your unit when a particular data file is changed.
```
odgrim, line 11: This feels unclear on when it is explicitly needed to use mark_as_read because we also add dag edges during include and dependency blocks.

docs/2.0/docs/pipelines/guides/file-dependencies.md
```diff
@@ -0,0 +1,15 @@
...
+
+However, there are some cases where you may want to run Pipelines for changes to files that are not part of your IaC. For example, you may have a `terragrunt.hcl` file that imports data from a `tags.yaml` file. If you make a change to the `tags.yaml` file, you may want to run Pipelines to plan/apply the changes to the associated `terragrunt.hcl` file.  This page documents how to configure pipelines to automatically run `plan/apply` when only these data files are changed.
+
+## Configuration
+
+Pipelines will automatically detect, based on git commit history, when data files have changed and emit a `FileChanged` event. This will create a new job in your workflow that will invoke Terragrunt with the [`queue-include-units-reading`](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-queue-include-units-reading) option set to the files that have changed. Terragrunt will then scan your code for units that depend on the changed files and add them to the plan/apply.
```
josh-padnick: What does "emit a `FileChanged` event" actually mean? What exactly emits that event? How is it picked up? Likewise, when you say "new job" and "workflow", what do you mean?

I realize that's not the focus of these docs, but as someone not super familiar with the technical internals of Pipelines, I was thrown off a little by this.

Also, I'm realizing that we're using some GitHub Actions-specific language here (job, workflow). Should we define general terms for these, and link to the official definitions so that users can know what we mean? Or perhaps say "GitHub Actions job" and "GitHub Actions workflow"?

docs/2.0/reference/pipelines/configurations.md
```diff
@@ -194,6 +178,25 @@ pipelines:
...
+
+This option is deprecated in favor of setting `TERRAGRUNT_PROVIDER_CACHE` as an [`env`](#env) option.
+</HclListItemDescription>
+<HclListItemDefaultValue defaultValue="false"/>
+<HclListItemExample>
+
+```yaml
+pipelines:
+  enable-terragrunt-provider-cache: true
+```
```
josh-padnick, line 195: It's really awkward that we have _two_ config languages, one "legacy" and one new, but only add values to the legacy one. Why have the new one at all?

Can we add these values to both the new and legacy config going forward, and reference both in all docs?

PR #2469, opened 2025-03-12
Revive "Component Modules"

josh-padnick:
The previous docs edits changed "building block modules" to the generic (and very confusing) "modules", removing virtually all meaning from this page.

This change brings back the distinct concept of a "building block" module, but since we're here, I've renamed it to a term @ebeneliason  and I liked better: "Component modules".

@ZachGoldberg: You'd be a good candidate to review this since you're up to speed on docs.

Oh, I updated this because I was about to post a comment on Reddit referencing the pages here, and then I noticed that this page had been changed in a confusing way!

## Summary by CodeRabbit

- **Documentation**
  - Reorganized and renamed content to focus on "Component Modules" and "Composite Modules" for improved clarity.
  - Updated introductory sections to explicitly define and emphasize the reusability of modules.
  - Expanded example descriptions to better illustrate functionalities of component and composite modules.
  - Revised terminology for alignment, transitioning from "building block modules" to "component modules" and from "Service Modules" to "Composite Modules."
  - Clarified descriptions and examples to reflect the new terminology consistently.
  - Introduced new documents detailing "Component Modules" and "Composite Modules."
  - Added new redirects to enhance navigation for updated documentation paths.


ZachGoldberg:
FYI - renaming files and changing paths will break old links, we should add links to `redirects.js` to ensure we don't cause unnecessary 404s

docs/2.0/docs/library/concepts/component-modules.md
```diff
@@ -0,0 +1,52 @@
+# Component Modules
```
ZachGoldberg, line 1: I don't love `component module` personally - What is a component?  It's a component of what?  Adjacent to the word "composite" it's a little clearer, but on its own I think the prior term "building block" was more easily understood.

Also - changing names is hard, not something we should do often, and we should make sure to overcommunicate with the rest of the team first.

ZachGoldberg (changes requested):
1) Needs redirects if we're changing URLs.
2) I'm personally not a fan of the rename, but if you feel strongly then maybe lets figure out in cafe?
3) The content as it was did actually still call things "Building Blocks", so I'm a bit unsure what the original bug is that inspired the PR

josh-padnick:
1. Redirects added.
2. I'm not too strong on "component modules." Might be a quick resolve in Cafe. [Discussion topic added](https://www.notion.so/gruntwork/Daily-Caf-50ca9ea355f042dab3c9d234df7d8baf?p=1b5d8759a682804a85f4e3660e97303f&pm=s).
3. "Building block" modules is used clearly in https://docs.gruntwork.io/2.0/docs/library/concepts/overview, but the [original building block modules page](https://docs.gruntwork.io/2.0/docs/library/concepts/modules) has been reverted to just "modules," which is pretty confusing as it stands.

ZachGoldberg:
> 3. "Building block" modules is used clearly in https://docs.gruntwork.io/2.0/docs/library/concepts/overview, but the [original building block modules page](https://docs.gruntwork.io/2.0/docs/library/concepts/modules) has been reverted to just "modules," which is pretty confusing as it stands.

I suppose I'm just confused about the "reverting" - I want to understand if this was an intentional change by us during docs project, or something done by AI during the editing phase.  Looking at the git history I'm not sure I see the revert you're talking about.

josh-padnick:
Spent a couple mins looking, and maybe it was in https://github.com/gruntwork-io/docs/pull/2093, the massive PR? Not sure either, but I assumed it was an Edify mistaken change. Originally, the "Modules" page was titled "Building Block Modules," whereas today both the title and content of that page are referred to only as "Modules."

ZachGoldberg:
> Spent a couple mins looking, and maybe it was in #2093, the massive PR? Not sure either, but I assumed it was an Edify mistaken change. Originally, the "Modules" page was titled "Building Block Modules," whereas today both the title and content of that page are referred to only as "Modules."

Looks like pre-docs project the page was called "modules" and the h1 heading was  "What is a module" 
image

josh-padnick:
Ok, looked into it a couple more mins, and you're actually right. I assumed this was a regression, but it actually isn't. Looking with fresh eyes at https://docs.gruntwork.io/2.0/docs/library/concepts/modules, I think it's fine as is, so I'll close this PR. Sorry for the noise. I'd like my 20 mins back.

PR #2540, opened 2025-04-22
feat: Adding performance docs

yhakbar:


docs/2.0/docs/pipelines/guides/improving-performance.md
```diff
@@ -0,0 +1,41 @@
+# Improving Performance
+
+## Overview
+
+Overall, given that Pipelines is a CI/CD tool for running Terragrunt, optimizing performance of Terragrunt is usually the best way to improve the performance of Pipelines.
+
+To get a quick list of Terragrunt performance optimizations, see the [Terragrunt Performance Optimization Guide](https://terragrunt.gruntwork.io/docs/troubleshooting/performance/).
```
ZachGoldberg, line 7: Can/should we include a handful of recommended TG flags here?

yhakbar: We link to the TG docs that do that. It seems far more sustainable to just document TG performance improvements there, unless there's special considerations for Pipelines.

ZachGoldberg: I think we should be prioritizing "make our customers lives easy" over "make our lives easy."  If there are default TG flags that customers should use to make TG faster, we should have those enabled by pipelines by default (with optouts) and we should document what those are here. I don't personally think it's effective for our enterprise product to tell customers to go read docs elsewhere and "figure it out yourself"

PR #3215, opened 2026-07-10
Add hook filter docs

Resonance1584:
## Summary by CodeRabbit

* **Documentation**
  * Added documentation for an optional `filter` field on after-hooks, including how scoping works via path, environment, and label matching.
  * Clarified hook execution behavior when filters match only some units (run only matched units; skip when nothing matches).
  * Updated the Pipelines Configurations as Code API reference with a new top-level `annotation` block (label assignment based on nested `filter`), and revised/expanded `filter` documentation and examples/requirements across related blocks.


ebeneliason (approved):
LGTM

